PR #159, opened 2022-06-28
Fix crash with empty style attr

### LINKED ISSUE #158: `no-dupe-style-properties` parse error when empty
`@ota-meshi/svelte/no-dupe-style-properties` throws an error:
```sql
TypeError: Cannot read property 'range' of undefined
```

When the attribute is empty, for example:
```html
<div class="navbar-main" id="main-navbar" style="" >
```

ota-meshi:
fixes #158

### DIFF
--- a/src/rules/require-optimized-style-attribute.ts
+++ b/src/rules/require-optimized-style-attribute.ts
@@ -32,7 +32,7 @@ export default createRule("require-optimized-style-attribute", {
         })
       },
       SvelteAttribute(node) {
-        if (node.key.name !== "style") {
+        if (node.key.name !== "style" || !node.value.length) {
           return
         }
         const root = parseStyleAttributeValue(node, context)
--- a/src/utils/css-utils/style-attribute.ts
+++ b/src/utils/css-utils/style-attribute.ts
@@ -32,6 +32,9 @@ export function parseStyleAttributeValue(
     return cache.get(node) || null
   }
   cache.set(node, null)
+  if (!node.value.length) {
+    return null
+  }
   const startOffset = node.value[0].range[0]
   const sourceCode = context.getSourceCode()
   const cssCode = node.value.map((value) => sourceCode.getText(value)).join("")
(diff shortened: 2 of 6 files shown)


ota-meshi:
I will be renaming after releasing this patch.

src/rules/require-optimized-style-attribute.ts
```diff
@@ -32,7 +32,7 @@ export default createRule("require-optimized-style-attribute", {
         })
       },
       SvelteAttribute(node) {
-        if (node.key.name !== "style") {
+        if (node.key.name !== "style" || !node.value.length) {
```
JounQin, line 35: Can `node.value` be `null`? `<span style>`

tests/fixtures/rules/no-shorthand-style-property-overrides/valid/empty01-input.svelte
```diff
@@ -0,0 +1,4 @@
+<script>
+</script>
+
+<div style="">...</div>
```
JounQin, line 4:
```suggestion
<div style>...</div>
<div style="">...</div>
```

JounQin (changes requested):

src/rules/require-optimized-style-attribute.ts
JounQin:
```suggestion
        if (node.key.name !== "style" || !node.value?.length) {
```

src/utils/css-utils/style-attribute.ts
```diff
@@ -32,6 +32,9 @@ export function parseStyleAttributeValue(
     return cache.get(node) || null
   }
   cache.set(node, null)
+  if (!node.value.length) {
```
JounQin, line 35:
```suggestion
  if (!node.value?.length) {
```

JounQin (changes requested):

ota-meshi:
Thank you for checking this PR! I added a test for attributes without equal and modified the source code.

src/rules/require-optimized-style-attribute.ts
ota-meshi: > Can node.value be null? `<span style>`

Currently it can not be `null`, but it is different from JSX AST, so it may be a bug of AST generated by svelte-eslint-parser 🤔

JounQin (approved):

Issue #157, opened 2022-06-28
Do you want npm package `eslint-plugin-svelte`?


JounQin:
@ota-meshi 

It was an alias point to [eslint-plugin-svelte3](https://github.com/sveltejs/eslint-plugin-svelte3) at https://github.com/JounQin/eslint-plugin-svelte, I think it's much suitable for this project.

JounQin:
Done.

PR #160, opened 2022-06-29
Rename package to eslint-plugin-svelte

ota-meshi:
This PR rename the package from `@ota-meshi/eslint-plugin-svelte` to `eslint-plugin-svelte`.

There is the following remaining work.
- [x] Add migration guide.
- [x] Drop support for Node12. #161
- [x] Remove deprecated rules. #162

Related to #157


package.json
```diff
@@ -109,6 +112,7 @@
     "eslint-plugin-node-dependencies": "^0.8.0",
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-plugin-regexp": "^1.0.0",
+    "eslint-plugin-svelte": "file:.",
```
ota-meshi, line 115: After publishing, change it to refer to npm.

JounQin, line 115:
```suggestion
    "eslint-plugin-svelte": "link:."
```

A soft like should more suitable for this case, and we can preserve it always.

tests/fixtures/rules/no-unused-svelte-ignore/valid/.eslintrc.js
```diff
@@ -0,0 +1,12 @@
+"use strict"
+
+module.exports = {
+  overrides: [
+    {
+      files: ["style-lang07-input.svelte"],
+      rules: {
+        "prettier/prettier": "off",
```
JounQin, line 8: Is this expected or is there anything wrong with prettier's rule? And `svelte` should be added into https://github.com/prettier/eslint-plugin-prettier/blob/master/eslint-plugin-prettier.js#L215-L225 for this plugin.

JounQin: Besides, for `eslint-plugin-svelte3`, it can be added in `https://github.com/prettier/eslint-plugin-prettier/blob/master/eslint-plugin-prettier.js#L190`

ota-meshi: The linting seems to work, but the format seems to fail. (And the log is output)
I haven't explored this issue deeply, but it's probably because prettier-plugin-svelte doesn't handle unknown langIds well. I disabled the rule as a workaround because the logs were in the way when I checked if lint worked :sweat_smile: 

image

https://github.com/ota-meshi/eslint-plugin-svelte/runs/7110654990?check_suite_focus=true

Input:

```svelte
<div class="foo">
  <div class="bar" />
</div>

<!-- svelte-ignore css-unused-selector -->
<style lang="unknown">
  .foo
    height: 20px
    width: 20px
    .foo
      height: 10px
      width: 10px
</style>
```

package.json
ota-meshi: If we specify a directory in the `file:` URL, I think a symbolic link will be used.
However, it may be different depending on the OS. It seems that symbolic links are used, at least on MacOS.
`link:.` didn't seem to work on my local npm. Does it use a different package manager?

JounQin: A constant development environment is very useful and necessary IMO.

.github/workflows/NodeCI.yml
```diff
@@ -44,10 +46,10 @@ jobs:
           node-version: ${{ matrix.node }}
       - name: Install ESLint ${{ matrix.eslint }}
         run: |+
-          npm i -D eslint@${{ matrix.eslint }} --legacy-peer-deps
-          npx rimraf node_modules
+          yarn add -D eslint@${{ matrix.eslint }} --ignore-engines
+          yarn run rimraf node_modules
```
JounQin, line 50: It's ubuntu, I believe `rm -rf` is always available already?

yarn.lock
```diff
@@ -0,0 +1,4280 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
JounQin, line 1: Previously, `npm-lock.json` is not generated and committed, so may it should be added into `.gitignore`?

package.json
```diff
@@ -12,25 +12,27 @@
...
-    "clean": "rimraf .nyc_output dist coverage",
+    "clean": "rimraf .nyc_output lib coverage build .svelte-kit svelte.config-dist.js",
     "pretest:base": "cross-env DEBUG=eslint-plugin-svelte*",
     "test": "mocha --require ts-node/register \"tests/src/**/*.ts\" --reporter dot --timeout 60000",
-    "cover": "nyc --reporter=lcov npm run test",
+    "cover": "nyc --reporter=lcov yarn test",
     "debug": "mocha --require ts-node/register/transpile-only \"tests/src/**/*.ts\" --reporter dot --timeout 60000",
     "lint": "eslint .",
     "eslint-fix": "eslint . --fix",
-    "update": "ts-node --transpile-only ./tools/update.ts && npm run format-for-gen-file",
```
JounQin, line 26: We can replace `ts-node` with [`tsx`](https://github.com/esbuild-kit/tsx) which is much faster due to `esbuild`, of course it can be a separate PR. If you want I can raise a PR after this PR merged.

tests/fixtures/rules/no-unused-svelte-ignore/valid/.eslintrc.js
JounQin: I just added better support for `eslint-plugin-svelte` in [`eslint-plugin-prettier@4.2.0`](https://github.com/prettier/eslint-plugin-prettier/blob/master/CHANGELOG.md#420)!

ota-meshi:
Thank you for your review. I changed this PR. So please check it again.

package.json
ota-meshi: You're right in team development. It makes sense to make the consistent development environment for this project.
Previously, I was the only one maintaining the project, so I wanted to prioritize giving the package manager a free choice for anyone trying to open a pull request, but that's no longer necessary.

JounQin (approved):
Let's move on!

PR #165, opened 2022-06-30
perf: use esbuild-register as node loader to replace ts-node

JounQin:


ota-meshi:
This change looks good, but I can't get it to work, probably because of an npm incident right now. I'll try it later and then merge it.
https://status.npmjs.org/incidents/6wr25yb0b2dd

ota-meshi (approved):
Thank you for this change! Your changed script seems to work pretty well!

PR #178, opened 2022-07-06
Fix false positives for component in `prefer-class-directive` rule

### LINKED ISSUE #176: `prefer-class-directive` should be ignored on a component

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.

8.19.0

2.1.0

### What did you do?

As mentioned in https://github.com/ota-meshi/eslint-plugin-svelte/issues/173 a reminder to fix a similar issue for `prefer-class-directive`.

.

### What actually happened?

.

### Link to Minimal Reproducible Example

.

### Additional comments

_No response_

ota-meshi:
fixes #176

### DIFF
--- a/src/utils/ast-utils.ts
+++ b/src/utils/ast-utils.ts
@@ -109,15 +109,18 @@ export function isHTMLElementLike(
   | (SvAST.SvelteSpecialElement & {
       name: SvAST.SvelteName & { name: "svelte:element" }
     }) {
-  if (node.type === "SvelteElement") {
-    if (node.kind === "html") {
-      return true
-    }
-    if (node.kind === "special") {
-      return node.name.name === "svelte:element"
-    }
+  if (node.type !== "SvelteElement") {
+    return false
   }
-  return false
+
+  switch (node.kind) {
+    case "html":
+      return true
+    case "special":
+      return node.name.name === "svelte:element"
+    default:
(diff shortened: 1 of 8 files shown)


JounQin (approved):

src/utils/ast-utils.ts
```diff
@@ -98,6 +98,28 @@ export function needParentheses(
...
+  node:
+    | SvAST.SvelteElement
+    | SvAST.SvelteScriptElement
+    | SvAST.SvelteStyleElement,
+): node is
+  | SvAST.SvelteHTMLElement
+  | (SvAST.SvelteSpecialElement & {
+      name: SvAST.SvelteName & { name: "svelte:element" }
+    }) {
+  if (node.type === "SvelteElement") {
```
JounQin, line 112: How about `return` earlier?

```ts
if (node.type !== 'SvelteElement') {
  return false
}

if (node.kind === 'html') {
  return true
}
if (node.kind === 'special') {
  return node.name.name === 'svelte:element'
}
```

ota-meshi: That's good! I change to use it.

JounQin (approved):

ota-meshi:
Thank you for your review!

PR #180, opened 2022-07-06
feat: indent/indentScript option

### LINKED ISSUE #155: indent rule requires that <script> elements be indented
This is how I want to write our svelte components:

```svelte
<script>
const foo = "bar";
</script>
```

But if I enable the indent rule, it really wants me to do this:

```svelte
<script>
    const foo = "bar";
</script>
```

I can disable it using `{ ignoredNodes : [ "SvelteScriptElement" ] }` but it means that none of the indentation for any direct children will be checked, which is kind of a bummer. I'm sure this is a bit annoying of an ask, but would you be open to adding a new option where a user can define the level of indentation they want specifically for `<script>`s? It could default to `1`, so it wouldn't be a breaking change.

If you're open to this proposal I'd be thrilled to figure out how to send a PR for it.

tivac:
Allows for configuring whether the contents of `<script>` blocks should be indented a single tabstop or not. Defaults to `true` for back-compat.

Fixes #155

### DIFF
--- a/src/rules/indent-helpers/svelte.ts
+++ b/src/rules/indent-helpers/svelte.ts
@@ -23,7 +23,12 @@ export function defineVisitor(context: IndentContext): NodeListener {
     // ELEMENTS
     // ----------------------------------------------------------------------
     SvelteScriptElement(node: AST.SvelteScriptElement) {
-      offsets.setOffsetElementList(node.body, node.startTag, node.endTag, 1)
+      offsets.setOffsetElementList(
+        node.body,
+        node.startTag,
+        node.endTag,
+        options.indentScript ? 1 : 0,
+      )
     },
     SvelteStyleElement(node: AST.SvelteStyleElement) {
       node.children.forEach((n) => offsets.ignore(n))
@@ -38,7 +43,7 @@ export function defineVisitor(context: IndentContext): NodeListener {
             node.children.filter(isNotEmptyTextNode),
             node.startTag,
             node.endTag,
-            1,
+            options.indentScript ? 1 : 0,
           )
         }
       } else {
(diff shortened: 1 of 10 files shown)


src/rules/indent-helpers/svelte.ts
```diff
@@ -38,7 +43,7 @@ export function defineVisitor(context: IndentContext): NodeListener {
             node.children.filter(isNotEmptyTextNode),
             node.startTag,
             node.endTag,
-            1,
+            options.indentScript ? 1 : 0,
```
ota-meshi, line 46: I don't think this change is necessary. Did we need this change?

ota-meshi (changes requested):

ota-meshi:
Thank you for this PR! I made one comment.

src/rules/indent-helpers/svelte.ts
tivac: Unclear, I'll test without tomorrow and see what happens.

JounQin (changes requested):
A new config for `prettier` with this rule disabledd should be added IMO. @ota-meshi

ota-meshi:
Normally, the fixture should be formatted. However, stylistic rule invalid tests and option tests may ignore the format.
If we want to ignore settings completely, we need to set it for each rule, `valid` and` invalid` directory. Also, we may want to ignore the format only partly.
I think it easier to use the ignore comment than use ignore settings.

ota-meshi:
Did I misunderstand what you were saying? Are you talking about the shareable configuration provided by this plugin?

JounQin:
@ota-meshi I mean add a new config in this plugin: `extends: ['plugin:svlete/prettier']` like `jsonc` plugin?

JounQin:
> Did I misunderstand what you were saying? Are you talking about the shareable configuration provided by this plugin?

Exactly.

ota-meshi:
Yeah. I think we need a new shareable configuration, as you say. We can work on it with another PR.

ota-meshi (approved):
LGTM! Thank you!

PR #181, opened 2022-07-07
Chore: update `@sveltejs/kit`

ota-meshi:
This PR updates the version of `@sveltejs/kit`. Related to #179.  
The way `@sveltejs/kit` is configured has changed so much that the configuration file has changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "prepublishOnly": "yarn clean && yarn build",
     "pretest:base": "cross-env DEBUG=eslint-plugin-svelte*",
     "preversion": "yarn test && git add .",
-    "svelte-kit": "node --experimental-loader ./svelte-kit-import-hook.mjs node_modules/@sveltejs/kit/svelte-kit.js",
+    "svelte-kit": "node --experimental-loader ./svelte-kit-import-hook.mjs node_modules/vite/bin/vite.js --config vite.config.mjs",
     "test": "yarn mocha \"tests/src/**/*.ts\" --reporter dot --timeout 60000",
     "ts": "node -r esbuild-register",
     "update": "yarn ts ./tools/update.ts && yarn format-for-gen-file",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "prepublishOnly": "yarn clean && yarn build",
     "pretest:base": "cross-env DEBUG=eslint-plugin-svelte*",
     "preversion": "yarn test && git add .",
-    "svelte-kit": "node --experimental-loader ./svelte-kit-import-hook.mjs node_modules/@sveltejs/kit/svelte-kit.js",
+    "svelte-kit": "node --experimental-loader ./svelte-kit-import-hook.mjs node_modules/vite/bin/vite.js --config vite.config.mjs",
```
JounQin, line 46: Can we use `yarn ts` here? And `vite.config.ts`(or `.mts`).

JounQin (changes requested):

package.json
JounQin: Besides, a `vite` script with `yarn ts node_modules/vite/bin` should be added, and reuse `yarn vite` here.

ota-meshi: Thank you for letting me know! I didn't know that vite.config.mts was available.

JounQin: You know, Vite uses `esbuild` under the hood. 🥰

ota-meshi: Hmm. It didn't work. Seems confusing with ESM and CJS. Probably because tsconfig is configured for CJS.

JounQin: Fine, then maybe we've just found a bug of Vite. 😅

ota-meshi: It may be resolved in the next release 😄 
https://github.com/vitejs/vite/pull/8729

JounQin: Maybe we can use the beta version?

JounQin: Fair enough. 😅

PR #182, opened 2022-07-07
Add `plugin:svelte/prettier` config

ota-meshi:
This PR adds `plugin:svelte/prettier` config that turns off rules that conflict with prettier-plugin-svelte.


JounQin (approved):
You're on fire! 🔥

PR #186, opened 2022-07-12
feat: add `svelte/html-closing-bracket-spacing`

marekvospel:
Added `html-closing-bracket-spacing` rule, which can be seen in [`eslint-plugin-vue`](https://github.com/vuejs/eslint-plugin-vue)

## Related PRs

feat: add `svelte/html-self-closing` #190 
feat: add `svelte/no-spaces-around-equal-signs-in-attribute` #191

## `svelte/html-closing-bracket-spacing`
A simple rule for enforcing a space before closing a tag
```html
<div/>
<!-- gets fixed to -->
<div />
```

## Coverage
100% test coverage.

## ~~Notes & questions~~
* ~~This is my first time contributing to this project and creating eslint plugins in general, so I'm sorry if some parts of my code aren't as good.~~
* ~~How should I document the rules? Is there a script to automatically generate the documentation, or should I create a new markdown file inside `docs/rules`? (This information should probably be added to contributing section in README)~~


ota-meshi:
Thank you for this PR!
You need to add documents that describes the rules you added. After adding the files, run `yarn update` to update the header and footer.

marekvospel:
@ota-meshi I've added the docs, fixed the merge conflicts and both the docs build and lint should be passing now!

ota-meshi:
Thank you for adding the documentation! I'm busy this week so I'll probably check it next week.

marekvospel:
I had extra time today, so I have added `svelte/html-closing-bracket-spacing` rule!
```html
<div/>
# gets fixed to
<div />
```
It's similar to to `svelte/no-spaces-around-equal-signs-in-attribute `, but only has to check the end of the tag, so it's a bit simpler rule.

The test also has 100% coverage

I've also fixed some minor problems in my docs (such as since version being 0.2.2, which is already released, or mistyped `<div />`

docs/rules/html-self-closing.md
```diff
@@ -0,0 +1,82 @@
...
+
+```json
+{
+  "svelte/html-self-closing": [
+    "error",
+    {
+      "html": {
+        "void": "never", // or "always" or "any"
+        "normal": "always", // or "never" or "any"
+        "custom": "always" // or "never" or "any"
```
ota-meshi, line 59:
```suggestion
        "component": "always" // or "never" or "any"
```

docs/rules/html-self-closing.md
```diff
@@ -0,0 +1,82 @@
...
+      }
+    }
+  ]
+}
+```
+
+- `html.void` (`"never"` by default)... Style of HTML void elements
+- `html.component` (`"always"` by default)... Style of svelte components
+- `html.normal` (`"always"` by default)... Style of other elements
```
ota-meshi, line 68: `<svelte:self></svelte:self>` seems to be affected by the `normal` option, but I think they need to share the `component` option or have different option.

docs/rules/html-self-closing.md
```diff
@@ -0,0 +1,82 @@
...
+        "void": "never", // or "always" or "any"
+        "normal": "always", // or "never" or "any"
+        "custom": "always" // or "never" or "any"
+      }
+    }
+  ]
+}
+```
+
+- `html.void` (`"never"` by default)... Style of HTML void elements
```
ota-meshi, line 66: Could you make `"always"` the default to be the same as the default for prettier-plugin-svelte?

src/rules/html-closing-bracket-spacing.ts
```diff
@@ -0,0 +1,127 @@
+import { createRule } from "../utils"
+import type { AST } from "svelte-eslint-parser"
+
+export default createRule("html-closing-bracket-spacing", {
+  meta: {
+    docs: {
+      description: "Require or disallow a space before tag's closing brackets",
```
ota-meshi, line 7: The first letter should be lowercase to be consistent with existing rules.

```suggestion
      description: "require or disallow a space before tag's closing brackets",
```

src/rules/html-closing-bracket-spacing.ts
```diff
@@ -0,0 +1,127 @@
...
+        const selfClosing =
+          tagSrc.slice(
+            Math.max(tagSrc.length - 2, 0),
+            Math.max(tagSrc.length - 1, 0),
+          ) === "/"
+
+        const tagType =
+          node.type === "SvelteEndTag"
+            ? "endTag"
+            : selfClosing
```
ota-meshi, line 107: I think you can use `node.selfClosing`.

src/rules/html-closing-bracket-spacing.ts
```diff
@@ -0,0 +1,127 @@
...
+          }
+        },
+      })
+    }
+
+    return {
+      "SvelteStartTag, SvelteEndTag"(
+        node: AST.SvelteStartTag | AST.SvelteEndTag,
+      ) {
+        if (node.type === "SvelteEndTag" && options.endTag === "any") return
```
ota-meshi, line 95: I think we can use `node.selfClosing` to merge the judges with the L111.

```js
        const tagType =
          node.type === "SvelteEndTag"
            ? "endTag"
            : node.selfClosing
            ? "selfClosingTag"
            : "startTag"

        if (options[tagType] === "any") return
```

src/rules/html-closing-bracket-spacing.ts
```diff
@@ -0,0 +1,127 @@
...
+      selfClosingTag: "always",
+      ...ctx.options[0],
+    }
+    const src = ctx.getSourceCode()
+
+    /**
+     * Returns true if string contains newline characters
+     */
+    function containsNewline(string: string): boolean {
+      return /.*\n.*/s.test(string)
```
ota-meshi, line 49:
```suggestion
      return string.includes("\n")
```

src/rules/html-closing-bracket-spacing.ts
```diff
@@ -0,0 +1,127 @@
...
+            : selfClosing
+            ? "selfClosingTag"
+            : "startTag"
+
+        if (node.type === "SvelteStartTag" && options[tagType] === "any") return
+
+        const match = /(\s*)\/?>$/m.exec(tagSrc)
+        if (containsNewline(match![1])) return
+
+        if (options[tagType] === "always" && !containsWhitespace(match![1])) {
```
ota-meshi, line 116: I don't think we need to use containsWhitespace.

```suggestion
        if (options[tagType] === "always" && !(match![1])) {
```

src/rules/html-closing-bracket-spacing.ts
```diff
@@ -0,0 +1,127 @@
...
+          report(node, true)
+        } else if (
+          options[tagType] === "never" &&
+          containsWhitespace(match![1])
```
ota-meshi, line 120:
```suggestion
          match![1]
```

src/rules/html-closing-bracket-spacing.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Report
+     */
+    function report(
+      node: AST.SvelteStartTag | AST.SvelteEndTag,
+      shouldHave: boolean,
+    ) {
+      const tagSrc = src.getText(node)
+      const match = /(\s*)\/?>$/m.exec(tagSrc)
```
ota-meshi, line 67: Probably don't need the m flag.

```suggestion
      const match = /(\s*)\/?>$/.exec(tagSrc)
```

I think the following HTML will make the wrong decision.

```html
<a attr="
  >
">
```

src/rules/html-closing-bracket-spacing.ts
```diff
@@ -0,0 +1,127 @@
...
+        const tagType =
+          node.type === "SvelteEndTag"
+            ? "endTag"
+            : selfClosing
+            ? "selfClosingTag"
+            : "startTag"
+
+        if (node.type === "SvelteStartTag" && options[tagType] === "any") return
+
+        const match = /(\s*)\/?>$/m.exec(tagSrc)
```
ota-meshi, line 113:
```suggestion
        const match = /(\s*)\/?>$/.exec(tagSrc)
```

ota-meshi (changes requested):
Thank you for this PR!

I have change requests.
I've looked at the source code for one rule, but I'll check the rest after tomorrow.

ota-meshi:
If possible, can you separate the PR for each rule?
That way I can split the time to check the PR.

ota-meshi:
If you want to create new rules besides these, you may find it useful to use `yarn new new-rule-name`.

docs-svelte-kit/build-system/src/eslint.mjs
```diff
@@ -1,6 +1,6 @@
 /* eslint require-jsdoc:0 -- shim */
 
-import * as all from "../../../node_modules/eslint/lib/linter/linter.js"
+import * as all from "../../../node_modules/eslint/lib/linter/linter"
```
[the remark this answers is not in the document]
JounQin: Adding `.js` extensions would help to migrate to native ESM later, but of course it should be consistent all through the whole code base.

tests/fixtures/rules/html-end-tags/valid/test01-input.svelte
```diff
@@ -0,0 +1 @@
+<p>This should be valid</p>
```
ota-meshi, line 1: It looks like an unnecessary file.

docs-svelte-kit/build-system/src/eslint.mjs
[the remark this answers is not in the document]
marekvospel: mb, I've added this when linting the project for some reason 😄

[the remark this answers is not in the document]
ota-meshi: Yes. It seems that the linting config was broken once. I think it's okay to put it back now.

src/rules/html-closing-bracket-spacing.ts
ota-meshi: Hmmm.  I checked on the parser demo site and each seems to return the following results:

```html
<img /> <!-- "selfClosing": true -->
<img> <!-- "selfClosing": false -->
```

https://ota-meshi.github.io/svelte-eslint-parser/

Can you share the template you tried, as there may be other causes?

marekvospel: I'm not sure rn, I'll try to find what caused my issue

.gitignore
```diff
@@ -108,3 +108,6 @@ dist
 /svelte.config-dist.js
 /docs-svelte-kit/shim/eslint.mjs
 /docs-svelte-kit/shim/assert.mjs
+
+# Intellij IDES
+.idea
```
JounQin, line 113: Personally speaking, `IDE` things should be ignored with global `gitignore` file on your own machine.

tsconfig.json
```diff
@@ -18,6 +18,7 @@
     //     "*": ["typings/*"]
     // },
     "esModuleInterop": true,
+    "resolveJsonModule": true,
```
JounQin, line 21: This could be broken for ESM later (although I know it is transformed as commonjs), but it could be better to not adding such unstandard config.

src/rules/html-closing-bracket-spacing.ts
marekvospel: Replacing "/" check with `node.selfClosing` doesn't break any tests, and I couldn't find any edge case where `startTag.selfClosing` would be `false` even if it was a self closing tag, so I think it was me mistaking something while adding this rule.

docs/rules/html-self-closing.md
marekvospel: I've moved this to #190, as it is related to `svelte/html-self-closing` rule and might take me longer to resolve.

JounQin:
Hi @marekvospel Can you update this PR’s description align to its actual content?

src/rules/html-closing-bracket-spacing.ts
```diff
@@ -0,0 +1,109 @@
...
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
+    messages: {
+      expectedSpace: "Expected space before '>', but not found.",
+      unexpectedSpace: "Expected no space before '>', but found.",
+    },
+    fixable: "code",
```
ota-meshi, line 33: I think we can use `"whitespace"` because it only auto-fixes whitespace.

ota-meshi (changes requested):
Thank you for changing this PR. Also, thank you for splitting the PR.
I commented only one. Other than that, it looks good to me.

JounQin:
Personally thinking, `styling` rules should be extracted to another plugin like `eslint-plugin-svelte-styling` just like ESLint core has frozen their styling rules. We have prettier for `.svelte` or `.vue` files, although someone may not be using `prettier`, but another `eslint-plugin-svelte-styling` can help, so that `styling` rules can be excluded from this one.

ota-meshi:
Hmm. I'm not sure if the stylistic rules should be an another plugin.
For now, I don't think it's necessary. Because the stylistic rules of the ESLint core are not yet another plugin.
But later my thinking may change.

JounQin:
> Because the stylistic rules of the ESLint core are not yet another plugin.

ESLint core has frozen their styling rules, no new related rule will be added.

ota-meshi:
I know that, but plugin for Svelte doesn't freeze the stylistic rules.
The ESLint core stylistic rules were used by many users and enhanced by many contributors until they froze. However, this plugin is not yet known to users and there are not enough stylistic rules.

ota-meshi (approved):
LGTM!

PR #189, opened 2022-07-17
Docs: Fix playground width

baseballyama:
Now playground width is too narorow.
image

To fix this issue, we need to override the layout svelte file's CSS in some way.
However, I could not figure out how to do it because of the markdown involved.
So I used `:global` to solve this issue.

image

I could not find a way not to use `!important`.


ota-meshi (approved):
Thank you!

PR #191, opened 2022-07-21
feat: add `svelte/no-spaces-around-equal-signs-in-attribute`

marekvospel:
Moved to separate PR from ota-meshi/eslint-plugin-svelte#186

## `svelte/no-spaces-around-equal-signs-in-attribute`
Also a simple rule which automatically removes spaces around attribute equal signs
```html
<div class ="test" />
<!-- gets fixed to -->
<div class="test" />
```
I had difficulties with this rule, as `SvelteAttribute` has only key and value, but with some regex magic the rule should work.

## Coverage
100% test coverage.

### DIFF
--- a/src/rules/no-spaces-around-equal-signs-in-attribute.ts
+++ b/src/rules/no-spaces-around-equal-signs-in-attribute.ts
@@ -1,0 +1,67 @@
+import { createRule } from "../utils"
+import type { AST } from "svelte-eslint-parser"
+
+export default createRule("no-spaces-around-equal-signs-in-attribute", {
+  meta: {
+    docs: {
+      description: "disallow spaces around equal signs in attribute",
+      category: "Stylistic Issues",
+      recommended: false,
+      conflictWithPrettier: true,
+    },
+    schema: {},
+    fixable: "code",
+    messages: {
+      noSpaces: "Unexpected spaces found around equal signs.",
+    },
+    type: "layout",
+  },
+  create(ctx) {
+    const source = ctx.getSourceCode()
+
+    /**
+     * Returns source text between attribute key and value, and range of that source
+     */
+    function getAttrEq(node: AST.SvelteAttribute): [string, AST.Range] {
+      const attrSource = source.getText(node)
+      const keyRange = node.key.range
+      const index =
+        /[^\s=]/.exec(attrSource.slice(keyRange[1] - keyRange[0]))?.index ?? 0
+      const valueStart = keyRange[1] + index
+      const eqSource = attrSource.slice(
+        keyRange[1] - keyRange[0],
+        valueStart - keyRange[0],
+      )
+
+      return [eqSource, [keyRange[1], keyRange[1] + eqSource.length]]
+    }
+
+    /**
+     * Returns true if string contains whitespace characters
+     */
+    function containsWhitespace(string: string): boolean {
+      return /.*\s.*/s.test(string)
+    }
+
+    return {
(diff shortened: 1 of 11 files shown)


src/rules/no-spaces-around-equal-signs-in-attribute.ts
```diff
@@ -0,0 +1,67 @@
...
+export default createRule("no-spaces-around-equal-signs-in-attribute", {
+  meta: {
+    docs: {
+      description: "disallow spaces around equal signs in attribute",
+      category: "Stylistic Issues",
+      recommended: false,
+      conflictWithPrettier: true,
+    },
+    schema: {},
+    fixable: "code",
```
ota-meshi, line 13: I think we can use `"whitespace"`.

src/rules/no-spaces-around-equal-signs-in-attribute.ts
```diff
@@ -0,0 +1,67 @@
...
+      const keyRange = node.key.range
+      const index =
+        /[^\s=]/.exec(attrSource.slice(keyRange[1] - keyRange[0]))?.index ?? 0
+      const valueStart = keyRange[1] + index
+      const eqSource = attrSource.slice(
+        keyRange[1] - keyRange[0],
+        valueStart - keyRange[0],
+      )
+
+      return [eqSource, [keyRange[1], keyRange[1] + eqSource.length]]
```
ota-meshi, line 36: I think it's simpler to slice directly from `source.text`.
Also, I think we can change the regex and get the `eqSource` directly.

```suggestion
      const keyRange = node.key.range
      const eqSource = /^[\s=]*/u.exec(
        source.text.slice(keyRange[1], node.range[1]),
      )![0]
      const valueStart = keyRange[1] + eqSource.length
      return [eqSource, [keyRange[1], valueStart]]
```

src/rules/no-spaces-around-equal-signs-in-attribute.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Returns true if string contains whitespace characters
+     */
+    function containsWhitespace(string: string): boolean {
+      return /.*\s.*/s.test(string)
```
ota-meshi, line 43: I don't think we need dot all.

```suggestion
      return /\s/u.test(string)
```

src/rules/no-spaces-around-equal-signs-in-attribute.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    return {
+      SvelteAttribute(node: AST.SvelteAttribute) {
```
ota-meshi, line 47: I think we should also check SvelteDirective and SvelteStyleDirective.

```svelte
<input bind:value  =  {value} style:width   =   "100px">
```

ota-meshi (changes requested):
Thank you for this PR!
I wrote comments.

src/rules/no-spaces-around-equal-signs-in-attribute.ts
ota-meshi: I realized that we should also check SvelteSpecialDirective as well.

```svelte
<svelte:element this  =  {expression}/>
```

https://github.com/ota-meshi/svelte-eslint-parser/blob/main/docs/AST.md#directives

ota-meshi (approved):
LGTM! Thank you!

PR #196, opened 2022-07-23
Add `svelte/sort-attributes` rule

### LINKED ISSUE #150: Add attribute sorting rule
### Motivation

Attributes on Svelte components and html can get unwieldy and hard to parse if there are too many. This is especially true when there is a lot of directives involved. One of the main thing I appreciate in eslint-plugin-react is their [`jsx-sort-props`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md) rule. It makes even complex components easy to read. 

### Proposed solution

Add sort-attributes rule to enforce a sort order of attributes. Alphabetical order would probably suffice for most cases however some sort of predefined "natural" order would be best. Here I'm thinking of the `in:` and `out:` directive. I think most would keep this two attributes logically together. `jsx-sort-props` has accumulated a fair amount of these, e.g.: `on:` event listeners last for Svelte. 

A simple alphabetical ordering I usually maintain manually.

```svelte
<Checkbox
  bind:this={checkbox}
  class="text-md"
  class:blue={isSelected}
  {completed}
  {id}
  in:fly
  on:change={handleChange}
  out:fade
  use:action
  {...$$props}/>
```

--- 
Appreciate any thought on this!

ota-meshi:
This PR adds `svelte/sort-attributes` rule that enforce order of attributes.

close #150


docs/rules/sort-attributes.md
```diff
@@ -0,0 +1,245 @@
...
+<slot name="content" {def} {abc} data-foo />
+```
+
+<!-- prettier-ignore-end -->
+
+</ESLintCodeBlock>
+
+## :wrench: Options
+
+```json
```
JounQin, line 106:
```suggestion
```jsonc
```

tivac:
Looked at the docs, love where this is headed and very impressed how fast it came together!

Quick question about the config format, I assume this uses standard regex flags so the `/u` usage in the examples isn't always necessary unless you need to match extended Unicode characters in attributes?

ota-meshi:
Thank you for checking! I will proceed with the implementation as it is 😉.

The `u` flag is not required. I like to use the u flag as similar to strict mode. The `u` flag will find the wrong use of the regex and let me know as an error. For example, `new RegExp ('\\w{1,5', 'u')`

PR #201, opened 2022-07-28
add `svelte/no-extra-reactive-curlies` rule

tivac:
Porting over a simple rule from https://github.com/tivac/eslint-plugin-svelte/ as per the conversation in https://github.com/ota-meshi/eslint-plugin-svelte/issues/198

I haven't been able to test that the documentation looks right, `yarn svelte-kit` is doing some funny path stuff on my windows machine. 🤷🏻‍♂️


ota-meshi:
What do you think about rename the rule to `no-lone-reactive-blocks` or `no-extra-reactive-curlies`? I think it's similar to eslint's `no-lone-blocks` and `no-extra-parens`.
https://eslint.org/docs/latest/rules/no-lone-blocks
https://eslint.org/docs/latest/rules/no-extra-parens

ota-meshi:
I didn't explain it, use the `yarn docs:watch` command to preview the document site. However, I'm not sure if there is a problem with windows.

src/rules/no-unnecessary-reactive-curlies.ts
```diff
@@ -0,0 +1,50 @@
...
+        node: AST.SvelteReactiveStatement,
+      ) => {
+        const source = context.getSourceCode()
+
+        return context.report({
+          node,
+          loc: node.loc,
+          messageId: "unnecessaryCurlies",
+
+          fix(fixer) {
```
ota-meshi, line 32: I think it is necessary to stop the autofix in the edge case that may break the source code.

```svelte
<script>
    $: { let name = 'world'; }
    $: { function fn(){} }
    $: { class MyClass {} }
</script>
```

ota-meshi (changes requested):
Thank you for this PR!
I have some change requests.

docs/rules/no-unnecessary-reactive-curlies.md
```diff
@@ -0,0 +1,48 @@
...
+  }
+</script>
+```
+
+</ESLintCodeBlock>
+
+## :wrench: Options
+
+Nothing.
+
```
ota-meshi, line 40: What do you think about adding something like the following wording to help with the migration?

> ## :heart: Compatibility
> 
> This rule was taken from [@tivac/eslint-plugin-svelte].  
> This rule is compatible with `@tivac/svelte/reactive-curlies` rule.
> 
> [@tivac/eslint-plugin-svelte]: https://github.com/tivac/eslint-plugin-svelte/

```suggestion

## :heart: Compatibility

This rule was taken from [@tivac/eslint-plugin-svelte].  
This rule is compatible with `@tivac/svelte/reactive-curlies` rule.

[@tivac/eslint-plugin-svelte]: https://github.com/tivac/eslint-plugin-svelte/

```

src/rules/no-unnecessary-reactive-curlies.ts
```diff
@@ -0,0 +1,50 @@
...
+import { createRule } from "../utils"
+
+export default createRule("no-unnecessary-reactive-curlies", {
+  meta: {
+    docs: {
+      description:
+        "disallow wrapping single reactive statements in curly braces",
+      category: "Stylistic Issues",
+      recommended: false,
+      conflictWithPrettier: true,
```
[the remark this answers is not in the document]
tivac: I removed this, but TS complains and says `conflictWithPrettier` is required. Should I bring it back set to `false` instead?

src/rules/no-unnecessary-reactive-curlies.ts
```diff
@@ -0,0 +1,50 @@
...
+    schema: [],
+    messages: {
+      unnecessaryCurlies: `Do not wrap reactive statements in curly braces unless necessary.`,
+    },
+    type: "suggestion",
+  },
+  create(context) {
+    return {
+      [`SvelteReactiveStatement > BlockStatement[body.length=1]`]: (
+        node: AST.SvelteReactiveStatement,
```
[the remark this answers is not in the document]
tivac: 👍🏻

src/rules/no-unnecessary-reactive-curlies.ts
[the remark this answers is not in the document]
ota-meshi: I made a mistake 😅. We need to set `conflictWithPrettier: false` as you say.

src/rules/no-unnecessary-reactive-curlies.ts
tivac: Added a check for the body statement to ensure that it's an assignment expression.

tivac:
> I didn't explain it, use the `yarn docs:watch` command to preview the document site. However, I'm not sure if there is a problem with windows.

This is what I see as output

```
C:\Users\pcavi\Documents\GitHub\eslint-svelte>yarn docs:watch
yarn run v1.22.19
$ yarn svelte-kit dev
$ node --experimental-loader ./svelte-kit-import-hook.mjs node_modules/vite/bin/vite.js dev
(node:9368) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9368) DeprecationWarning: Obsolete loader hook(s) supplied and will be ignored: transformSource
build@ \C:\Users\pcavi\Documents\GitHub\eslint-svelte\docs-svelte-kit\build-system\src\eslint.mjs
X [ERROR] Could not read from file: C:\Users\pcavi\Documents\GitHub\eslint-svelte\C:\Users\pcavi\Documents\GitHub\eslint-svelte\docs-svelte-kit\build-system\src\process-shim.mjs

X [ERROR] Could not resolve "\\C:\\Users\\pcavi\\Documents\\GitHub\\eslint-svelte\\docs-svelte-kit\\build-system\\src\\eslint.mjs"

failed to load config from C:\Users\pcavi\Documents\GitHub\eslint-svelte\vite.config.mts
error when starting dev server:
Error: Build failed with 2 errors:
error: Could not read from file: C:\Users\pcavi\Documents\GitHub\eslint-svelte\C:\Users\pcavi\Documents\GitHub\eslint-svelte\docs-svelte-kit\build-system\src\process-shim.mjs
error: Could not resolve "\\C:\\Users\\pcavi\\Documents\\GitHub\\eslint-svelte\\docs-svelte-kit\\build-system\\src\\eslint.mjs"
    at failureErrorWithLog (C:\Users\pcavi\Documents\GitHub\eslint-svelte\node_modules\esbuild\lib\main.js:1621:15)
    at C:\Users\pcavi\Documents\GitHub\eslint-svelte\node_modules\esbuild\lib\main.js:1263:28
    at runOnEndCallbacks (C:\Users\pcavi\Documents\GitHub\eslint-svelte\node_modules\esbuild\lib\main.js:1176:65)
    at buildResponseToResult (C:\Users\pcavi\Documents\GitHub\eslint-svelte\node_modules\esbuild\lib\main.js:1261:7)
    at C:\Users\pcavi\Documents\GitHub\eslint-svelte\node_modules\esbuild\lib\main.js:1374:14
    at C:\Users\pcavi\Documents\GitHub\eslint-svelte\node_modules\esbuild\lib\main.js:675:9
    at handleIncomingPacket (C:\Users\pcavi\Documents\GitHub\eslint-svelte\node_modules\esbuild\lib\main.js:772:9)
    at Socket.readFromStdout (C:\Users\pcavi\Documents\GitHub\eslint-svelte\node_modules\esbuild\lib\main.js:641:7)
    at Socket.emit (node:events:527:28)
    at addChunk (node:internal/streams/readable:315:12)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

This one in particular seems like a likely culprit with the doubled-up folder structure.

`X [ERROR] Could not read from file: C:\Users\pcavi\Documents\GitHub\eslint-svelte\C:\Users\pcavi\Documents\GitHub\eslint-svelte\docs-svelte-kit\build-system\src\process-shim.mjs`

src/rules/no-extra-reactive-curlies.ts
```diff
@@ -0,0 +1,60 @@
...
+    return {
+      [`SvelteReactiveStatement > BlockStatement[body.length=1]`]: (
+        node: TSESTree.BlockStatement,
+      ) => {
+        // $: { foo = "bar"; }
+        // Only want to transform if the contents of the block is a single assignment
+        // Anything else gets us into potentially weird territory and probably isn't worth handling
+        if (
+          node.body[0].type !== "ExpressionStatement" ||
+          node.body[0].expression.type !== "AssignmentExpression"
```
ota-meshi, line 30: I would like to omit the curly braces in the reactive statement as follows, not just in the assignment:

```svelte
<script>
$: {
  if (foo) { // and `for`, `switch`, `while`, `try`, and more
    // ...
  }
}
$: {
  fn(foo, bar)
}
</script>
```

Do you think these should be covered by this rule?

ota-meshi (changes requested):

ota-meshi:
> This one in particular seems like a likely culprit with the doubled-up folder structure.

Hmm. I don't know how to fix it. I think we giving esbuild the correct absolute path.

src/rules/no-extra-reactive-curlies.ts
tivac: Ideally yes! The more I think about this the more I think the "invalid with braces removed" examples you provided are uncommon enough that we shouldn't try to account for them in this rule and let folks disable it if they want to do something as weird as defining a class in a reactive statement.

Otherwise no matter what it's going to end up being a chunk of conditional logic that is always gonna be hard to reason about in the rule, and I'm worried that either the whitelist or blacklist approach are both gonna be a maintenance hassle. Thoughts?

ota-meshi: I think it probably makes sense to use an allowed list approach.
The spec seems to allow statements. (However, since it is different from ESTree, the declaration is not included.)
https://tc39.es/ecma262/2022/multipage/ecmascript-language-statements-and-declarations.html#prod-LabelledItem

Given the specs and the actual method used, I think the complete list is below.

- BlockStatement
- ExpressionStatement
- IfStatement
- ForStatement
- ForInStatement
- ForOfStatement
- WhileStatement
- DoWhileStatement
- SwitchStatement
- TryStatement

The following is a code example:

```js
$: {} // BlockStatement
/* Defined as VariableStatement in ECMAScript. */
$: var a = "" // VariableDeclaration with var (not let and const)
$: ; // EmptyStatement
$: foo() // ExpressionStatement
$: if (foo) {} // IfStatement
/* Defined as BreakableStatement in ECMAScript. */
$: for(;;) {} // ForStatement
$: for(const a in lint) {} // ForInStatement
$: for(const a of lint) {} // ForOfStatement
$: while(foo) {} // WhileStatement
$: do {} while(foo) // DoWhileStatement
$: switch (a) { case 1: {} } // SwitchStatement
// $: continue // ContinueStatement (Syntax Error: not available at the top level.)
// $: break // BreakStatement (Syntax Error: not available at the top level.)
$: return // ReturnStatement ( globalReturn: true )
// $: with (foo) {} // WithStatement (Syntax Error: not available in strict mode.)
$: throw new Error() // ThrowStatement
$: try {} finally {} // TryStatement


/* Defined as LexicalDeclaration in ECMAScript. */
// $: const b = "" // Syntax Error
// $: let c = "" // Syntax Error
/* Defined as ClassDeclaration in ECMAScript. */
// $: class MyClass {} // Syntax Error
/* Defined as HoistableDeclaration in ECMAScript. */
// $: function fn(){} // Syntax Error (strict mode)
```

ota-meshi: I forgot to include the LabeledStatement 😓 

```js
$: foo: for (;;) {} // LabeledStatement
```

tivac: I can look into supporting all of the various possibilities but I'm not particularly excited about it. There's more that are allowed than disallowed, so I think a blacklist makes more sense. But even paring it down by doing a blacklist it's hard to justify supporting stuff like not transforming a reactive class declaration (wouldn't that throw an error about class redeclaration?) or a reactive`const`? The only way that makes any sense is with more than one statement because you can't access the `const` outside of the reactive statement.

I'm not really sure where to take this rule because either approach feels like a massive explosion of complexity for very small returns. Maybe instead of a fixable rule it just offers a suggestion and we keep it simpler?

ota-meshi: > Maybe instead of a fixable rule it just offers a suggestion and we keep it simpler?

Yeah. I think that is a good idea :+1:

ota-meshi (approved):
LGTM! Thank you very much

PR #202, opened 2022-07-29
Fix false positives for files other than `*.svelte` in `svelte/no-unused-svelte-ignore` rule

ota-meshi:
This PR fixes a false positive in the `svelte/no-unused-svelte-ignore` rule for JS files.
The rule should not check for files other than those parsed by svelte-eslint-parser.

I also changed the test utility so that it can target fixtures other than .svelte.

### DIFF
--- a/src/rules/no-unused-svelte-ignore.ts
+++ b/src/rules/no-unused-svelte-ignore.ts
@@ -19,6 +19,9 @@ export default createRule("no-unused-svelte-ignore", {
   },
 
   create(context) {
+    if (!context.parserServices.isSvelte) {
+      return {}
+    }
     const sourceCode = context.getSourceCode()
 
     const ignoreComments: IgnoreItem[] = []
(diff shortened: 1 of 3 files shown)


tests/utils/utils.ts
```diff
@@ -222,23 +222,20 @@ function getConfig(ruleName: string, inputFile: string) {
...
     configFile = path.join(path.dirname(inputFile), "_config.json")
   }
   if (fs.existsSync(configFile)) {
     config = JSON.parse(fs.readFileSync(configFile, "utf8"))
   }
+  const parser =
+    path.extname(filename) === ".svelte"
+      ? require.resolve("svelte-eslint-parser")
+      : undefined
   if (config && typeof config === "object") {
```
JounQin, line 236: Is this check necessary? Shouldn't `return Object.assign({ parser }, config, { code, filename })` always work?

src/rules/no-unused-svelte-ignore.ts
```diff
@@ -19,6 +19,9 @@ export default createRule("no-unused-svelte-ignore", {
   },
 
   create(context) {
+    if (!context.parserServices.isSvelte) {
```
JounQin, line 22: 👍

tests/utils/utils.ts
ota-meshi: Thank you for let me know. I didn't notice it. I fixed it.

JounQin (approved):

PR #203, opened 2022-07-29
Add `svelte/no-reactive-literals` rule

tivac:
Porting over another of the simpler rules from https://github.com/tivac/eslint-plugin-svelte/ like we discussed in https://github.com/ota-meshi/eslint-plugin-svelte/issues/198

### DIFF
--- a/docs/rules/no-reactive-literals.md
+++ b/docs/rules/no-reactive-literals.md
@@ -1,0 +1,32 @@
+#  (svelte/no-reactive-literals)
+
+> Don't assign literal values in reactive statements
+
+## :book: Rule Details
+
+This rule reports on any assignment of a static, unchanging value within a reactive statement because it's not necessary.
+
+<ESLintCodeBlock>
+
+<!--eslint-skip-->
+
+```svelte
+<script>
+  /* eslint svelte/no-reactive-literals: "error" */
+  /* ✓ GOOD */
+  let foo = "bar";
+  
+  /* ✗ BAD */
+  $: foo = "bar";
+</script>
+```
+</ESLintCodeBlock>
+
+## :wrench: Options
+
+Nothing
+
+## :mag: Implementation
+
+- [Rule source](https://github.com/ota-meshi/eslint-plugin-svelte/blob/main/src/rules/no-reactive-literals.ts)
+- [Test source](https://github.com/ota-meshi/eslint-plugin-svelte/blob/main/tests/src/rules/no-reactive-literals.ts)
--- a/docs/rules/no-reactive-literals.md
+++ b/docs/rules/no-reactive-literals.md
@@ -1,6 +1,15 @@
-#  (svelte/no-reactive-literals)
+---
+pageClass: "rule-details"
+sidebarDepth: 0
+title: "svelte/no-reactive-literals"
+description: "Don't assign literal values in reactive statements"
+---
+
+# svelte/no-reactive-literals
 
 > Don't assign literal values in reactive statements
+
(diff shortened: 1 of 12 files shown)


src/rules/no-reactive-literals.ts
```diff
@@ -0,0 +1,69 @@
...
+      // $: foo = 1;
+      [`${labeledStatementBase}[right.type="Literal"]`]: warn,
+
+      // $: foo = [];
+      [`${labeledStatementBase}[right.type="ArrayExpression"][right.elements.length=0]`]:
+        warn,
+
+      // $: foo = {};
+      [`${labeledStatementBase}[right.type="ObjectExpression"][right.properties.length=0]`]:
+        warn,
```
ota-meshi, line 66: I think you can make it simpler by using `:matches()`.

```suggestion
      [`SvelteReactiveStatement > ExpressionStatement > AssignmentExpression:matches(${[
        // $: foo = "foo";
        // $: foo = 1;
        `[right.type="Literal"]`,
        // $: foo = [];
        `[right.type="ArrayExpression"][right.elements.length=0]`,
        // $: foo = {};
        `[right.type="ObjectExpression"][right.properties.length=0]`,
      ].join(",")})`]: warn,
```

https://github.com/estools/esquery#:~:text=matches%2Dany%3A-,%3Amatches,-(%5Battr%5D%20%3E%20%3Afirst%2Dchild

src/rules/no-reactive-literals.ts
```diff
@@ -0,0 +1,69 @@
...
+
+      return context.report({
+        node: parent,
+        loc: parent.loc,
+        messageId: "noReactiveLiterals",
+
+        fix(fixer) {
+          return [
+            // Insert "let" + whatever was in there
+            fixer.insertTextBefore(parent, `let ${source.getText(node)}`),
```
ota-meshi, line 46: I think it's a difficult to make auto-fixes. If the variable has already been declared, you will get a syntax error.

```svelte
<script>
let foo;
$: foo = 42;
</script>
```

We need to check the predefined variables for auto-fix. Or I think it is good not to auto-fix.

src/rules/no-reactive-literals.ts
```diff
@@ -0,0 +1,69 @@
+import type { TSESTree } from "@typescript-eslint/types"
+import { createRule } from "../utils"
+
+const labeledStatementBase = `SvelteReactiveStatement > ExpressionStatement > AssignmentExpression`
+
+export default createRule("no-reactive-literals", {
+  meta: {
+    docs: {
+      description: "Don't assign literal values in reactive statements",
+      category: "Stylistic Issues",
```
ota-meshi, line 10: Probably not a stylistic issue. I think this rule should be `"Best Practices"`.

```suggestion
      category: "Best Practices",
```

ota-meshi (changes requested):

src/rules/no-reactive-literals.ts
tivac: Swapped to suggestions.

tivac:
Had to add some infrastructure to make snapshot testing of suggestions possible, will be useful for #201 as well.

docs/rules/no-reactive-literals.md
```diff
@@ -0,0 +1,32 @@
+#  (svelte/no-reactive-literals)
```
ota-meshi, line 1: Could you run `yarn update` to regenerate the documentation headers?

ota-meshi (approved):
LGTM! Thank you!

PR #205, opened 2022-07-31
Add `prefer-reactive-destructuring` rule

tivac:
From the docs:

----

This rule triggers whenever a reactive statement contains an assignment that could be rewritten using destructuring. This is beneficial because it allows svelte's change-tracking to prevent wasteful redraws whenever the object is changed.

This [svelte REPL](https://svelte.dev/repl/96759f4772314d0a840e11370ef76711) example shows just how effective this can be at preventing bogus redraws.

```svelte
<script>
  /* eslint svelte/prefer-reactive-destructuring: "error" */
  /* ✓ GOOD */
  $: ({ foo } = info)
  $: ({ bar: baz } = info)

  /* ✗ BAD */
  $: foo = info.foo
  $: baz = info.bar
</script>
```

### DIFF
--- a/src/rules/prefer-reactive-destructuring.ts
+++ b/src/rules/prefer-reactive-destructuring.ts
@@ -40,23 +40,27 @@ export default createRule("prefer-reactive-destructuring", {
           node,
           loc: node.loc,
           messageId: "useDestructuring",
-          suggest: [
-            {
-              messageId: "suggestDestructuring",
-              fix: (fixer) => [
-                fixer.insertTextBefore(
-                  lToken,
-                  matched ? `({ ` : `({ ${prop}: `,
-                ),
-                fixer.insertTextAfter(lToken, ` }`),
-                fixer.replaceTextRange(
-                  [rTokens[0].range[0], rTokens[1].range[1]],
-                  "",
-                ),
-                fixer.insertTextAfter(rTokens[1], ")"),
-              ],
-            },
-          ],
+          suggest:
+            // Don't show suggestions for entries like $: info = foo.bar.info, the destructuring
+            // just looks too gross and complicates the rule too much
+            right.object.type === "MemberExpression"
+              ? []
+              : [
+                  {
+                    messageId: "suggestDestructuring",
+                    fix: (fixer) => [
+                      fixer.insertTextBefore(
+                        lToken,
+                        matched ? `({ ` : `({ ${prop}: `,
+                      ),
+                      fixer.insertTextAfter(lToken, ` }`),
+                      fixer.replaceTextRange(
+                        [rTokens[0].range[0], rTokens[1].range[1]],
+                        ")",
+                      ),
+                    ],
+                  },
+                ],
         })
       },
     }
--- a/src/rules/prefer-reactive-destructuring.ts
+++ b/src/rules/prefer-reactive-destructuring.ts
@@ -41,9 +41,8 @@ export default createRule("prefer-reactive-destructuring", {
           loc: node.loc,
           messageId: "useDestructuring",
           suggest:
-            // Don't show suggestions for entries like $: info = foo.bar.info, the destructuring
-            // just looks too gross and complicates the rule too much
-            right.object.type === "MemberExpression"
+            // Don't show suggestions for complex right-hand values, too tricky to get it right
+            right.object.type !== "Identifier" || right.computed
               ? []
               : [
                   {
--- a/src/rules/prefer-reactive-destructuring.ts
+++ b/src/rules/prefer-reactive-destructuring.ts
@@ -1,0 +1,64 @@
+import type { TSESTree } from "@typescript-eslint/types"
+import { createRule } from "../utils"
+
+export default createRule("prefer-reactive-destructuring", {
+  meta: {
+    docs: {
+      description: "Prefer destructuring from objects in reactive statements",
+      category: "Best Practices",
+      recommended: false,
+    },
+    hasSuggestions: true,
+    schema: [],
+    messages: {
+      useDestructuring: `Prefer destructuring in reactive statements`,
+      suggestDestructuring: `Use destructuring to get finer-grained redraws`,
+    },
+    type: "suggestion",
+  },
+  create(context) {
+    return {
+      // Finds: $: info = foo.info
+      // Suggests:  $: ({ info } = foo);
+      [`SvelteReactiveStatement > ExpressionStatement > AssignmentExpression[left.type="Identifier"][right.type="MemberExpression"]`](
+        node: TSESTree.AssignmentExpression,
+      ) {
+        const left = node.left as TSESTree.Identifier
+        const right = node.right as TSESTree.MemberExpression
+
+        const prop = (right.property as TSESTree.Identifier).name
+
+        const source = context.getSourceCode()
+        const lTokens = source.getTokens(left)
+        const rTokens = source.getTokens(right)
+        const matched = prop === left.name
+
+        return context.report({
+          node,
+          loc: node.loc,
+          messageId: "useDestructuring",
+          suggest: [
+            {
+              messageId: "suggestDestructuring",
+              fix: (fixer) => [
+                fixer.insertTextBefore(
+                  lTokens[0],
+                  matched ? `({ ` : `({ ${prop}: `,
+                ),
+                fixer.insertTextAfter(lTokens[0], ` }`),
+                fixer.replaceTextRange(
+                  [
+                    rTokens[rTokens.length - 2].range[0],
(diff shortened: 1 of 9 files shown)


src/rules/prefer-reactive-destructuring.ts
```diff
@@ -0,0 +1,64 @@
...
+      [`SvelteReactiveStatement > ExpressionStatement > AssignmentExpression[left.type="Identifier"][right.type="MemberExpression"]`](
+        node: TSESTree.AssignmentExpression,
+      ) {
+        const left = node.left as TSESTree.Identifier
+        const right = node.right as TSESTree.MemberExpression
+
+        const prop = (right.property as TSESTree.Identifier).name
+
+        const source = context.getSourceCode()
+        const lTokens = source.getTokens(left)
```
ota-meshi, line 32: I think we can use `getFirstToken` if we only use the first token.

src/rules/prefer-reactive-destructuring.ts
```diff
@@ -0,0 +1,64 @@
...
+        const rTokens = source.getTokens(right)
```
ota-meshi, line 33: I think we can probably use getLastTokens. We can specify how many to get with the count option.

src/rules/prefer-reactive-destructuring.ts
```diff
@@ -0,0 +1,64 @@
...
+                ),
+                fixer.insertTextAfter(lTokens[0], ` }`),
+                fixer.replaceTextRange(
+                  [
+                    rTokens[rTokens.length - 2].range[0],
+                    rTokens[rTokens.length - 1].range[1],
+                  ],
+                  "",
+                ),
+                fixer.insertTextAfter(rTokens[rTokens.length - 1], ")"),
```
ota-meshi, line 56: We can probably change it to:

```suggestion
                  ")",
                ),
```

src/rules/prefer-reactive-destructuring.ts
```diff
@@ -0,0 +1,64 @@
...
+      useDestructuring: `Prefer destructuring in reactive statements`,
+      suggestDestructuring: `Use destructuring to get finer-grained redraws`,
+    },
+    type: "suggestion",
+  },
+  create(context) {
+    return {
+      // Finds: $: info = foo.info
+      // Suggests:  $: ({ info } = foo);
+      [`SvelteReactiveStatement > ExpressionStatement > AssignmentExpression[left.type="Identifier"][right.type="MemberExpression"]`](
```
ota-meshi, line 23: Isn't member access like the following checked by this rule? Doesn't it affect the number of redraws in the first place?

```svelte
$: infoBar = foo.info.bar
```

ota-meshi (changes requested):

src/rules/prefer-reactive-destructuring.ts
tivac: Updated [the REPL](https://svelte.dev/repl/96759f4772314d0a840e11370ef76711) to show that `$: infoBar = foo.info.bar` style reactive statements will still re-render every time `foo` changes.

src/rules/prefer-reactive-destructuring.ts
tivac: `fix()` returns have to be fixer objects, as far as I can tell it doesn't allow bare strings to be returned.

ota-meshi: I don't mean bare strings. I thought replaceTextRange and insertTextAfter could be merged into one replaceTextRange.

src/rules/prefer-reactive-destructuring.ts
ota-meshi: Are you saying that changing to the following will reduce re-renders?

```js
$: ({ bar: infoBar } = foo.info)
```

If so, could you please add the following test case? I don't think the current selector checks that.

```js
$: infoBar = foo.info.bar
```

src/rules/prefer-reactive-destructuring.ts
```diff
@@ -0,0 +1,64 @@
...
+                fixer.insertTextBefore(
+                  lToken,
+                  matched ? `({ ` : `({ ${prop}: `,
+                ),
+                fixer.insertTextAfter(lToken, ` }`),
+                fixer.replaceTextRange(
+                  [rTokens[0].range[0], rTokens[1].range[1]],
+                  "",
+                ),
+                fixer.insertTextAfter(rTokens[1], ")"),
```
ota-meshi, line 56: See https://github.com/ota-meshi/eslint-plugin-svelte/pull/205#discussion_r934992304

src/rules/prefer-reactive-destructuring.ts
tivac: Ah, whoops. Yeah easy change.

src/rules/prefer-reactive-destructuring.ts
tivac: I'm going to have it bail on multiply-nested rules and not make any suggestions, it's just too gross-looking. If the user wants to do the multiple-levels of destructuring more power to 'em.

ota-meshi: Thanks for adding the test cases.

> I don't think the current selector checks that.

I misread the selector 😓. It was already checked.

src/rules/prefer-reactive-destructuring.ts
```diff
@@ -0,0 +1,68 @@
...
+        const matched = prop === left.name
+
+        return context.report({
+          node,
+          loc: node.loc,
+          messageId: "useDestructuring",
+          suggest:
+            // Don't show suggestions for entries like $: info = foo.bar.info, the destructuring
+            // just looks too gross and complicates the rule too much
+            right.object.type === "MemberExpression"
```
ota-meshi, line 46: I just tested this rule and it seems there is a bug in the suggestion.

```svelte
<script>
  /* eslint svelte/prefer-reactive-destructuring: "error" */
  $: foo = info[foo]
  $: baz = info["bar"]
</script>
```

Applying the suggestions:

```svelte
<script>
  /* eslint svelte/prefer-reactive-destructuring: "error" */
  $: ({ foo } = info[)
  $: ({ undefined: baz } = info[)
</script>
```

ota-meshi (changes requested):

src/rules/prefer-reactive-destructuring.ts
ota-meshi, line 46: The following code makes a suggestion, is that intentional?

```svelte
<script>
  /* eslint svelte/prefer-reactive-destructuring: "error" */
  $: foo = info.foo().bar
</script>
```

If it's not intentional, I think you should do something like this:

```suggestion
            right.object.type !== "Identifier"
```

ota-meshi: I think it makes sense to stop suggesting if `right.computed` is true.

https://github.com/estree/estree/blob/master/es5.md#memberexpression

ota-meshi:
I've noticed that sometimes what the rule report doesn't lead to a reduction in render times.
I think the documentation might need to be fixed.

See https://github.com/ota-meshi/eslint-plugin-svelte/pull/208#discussion_r942125022

src/rules/prefer-reactive-destructuring.ts
```diff
@@ -0,0 +1,67 @@
+import type { TSESTree } from "@typescript-eslint/types"
+import { createRule } from "../utils"
+
+export default createRule("prefer-reactive-destructuring", {
+  meta: {
+    docs: {
+      description: "Prefer destructuring from objects in reactive statements",
```
marekvospel, line 7: The first letter should be lowercase to be consistent with existing rules. (see #218)

baseballyama:
@tivac What are the remaining tasks of the PR? I can do the remaining tasks.

tivac:
@baseballyama Needs to be changed to a stylistic rule and any remaining comments need to be addressed (I think it's just capitalization nit?). Seems like either svelte has improved object access change tracking or my initial tests when I wrote these rules years ago were faulty 🤷🏻‍♂️ So this is really a style preference over a perf thing.

baseballyama:
I checked this rule well, and now I understand this rule doesn't have an impact on rendering frequency.
And already ESLint can report destructuring through `prefer-destructuring` rule.

So I think even in terms of styling, I don't think we need to add this rule.
I would like to clarify the needs of this rule.

ota-meshi:
Thank you for the explanation.

> And already ESLint can report destructuring through prefer-destructuring rule.
> So I think even in terms of styling, I don't think we need to add this rule.

You are right. So I will close this PR.

PR #206, opened 2022-07-31
add `no-reactive-functions` rule

tivac:
## :book: Rule Details

This rule reports whenever a function is defined in a reactive statement. This isn't necessary, as each time the function is executed it'll already have access to the latest values necessary. Redefining the function in the reactive statement is just a waste of CPU cycles.

```svelte
<script>
  /* eslint svelte/no-reactive-functions: "error" */

  /* ✓ GOOD */
  const arrowFn = () => { /* ... */ }
  const func = function() { /* ... */ }
  
  /* ✗ BAD */
  $: arrowFn = () => { /* ... */ }
  $: func = function() { /* ... */ }
</script>
```

### DIFF
--- a/docs/rules/no-reactive-functions.md
+++ b/docs/rules/no-reactive-functions.md
@@ -1,0 +1,51 @@
+---
+pageClass: "rule-details"
+sidebarDepth: 0
+title: "svelte/no-reactive-functions"
+description: ""
+---
+
+# svelte/no-reactive-functions
+
+> Don't assign functions in reactive statements
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> **_This rule has not been released yet._** </badge>
+
+## :book: Rule Details
+
+This rule reports whenever a function is defined in a reactive statement. This isn't necessary, as each time the function is executed it'll already have access to the latest values necessary. Redefining the function in the reactive statement is just a waste of CPU cycles.
+
+<ESLintCodeBlock>
+
+<!--eslint-skip-->
+
+```svelte
+<script>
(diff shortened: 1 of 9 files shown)


src/rules/no-reactive-functions.ts
```diff
@@ -0,0 +1,56 @@
...
+            {
+              messageId: "fixReactiveFns",
+              fix(fixer) {
+                const tokens = source.getTokens(parent)
+
+                // Replace the entire reactive label with "const"
+                return fixer.replaceTextRange(
+                  [tokens[0].range[0], tokens[1].range[1]],
+                  "const",
+                )
```
ota-meshi, line 48: If there is no space after the colon (`:`), I think we should change it to:

```suggestion
                // Replace the entire reactive label with "const"
                return fixer.replaceTextRange(
                  [parent.range[0], node.parent.parent.range[0],
                  "const ",
                )
```

Test case
```js
/* prettier-ignore */
$:bar = foo.bar
```

ota-meshi (changes requested):

docs/rules/no-reactive-functions.md
```diff
@@ -0,0 +1,51 @@
+---
+pageClass: "rule-details"
+sidebarDepth: 0
+title: "svelte/no-reactive-functions"
+description: ""
```
ota-meshi, line 5: Could you run `yarn update` and update the header?

ota-meshi (approved):
LGTM! Thank you very much!

JounQin (approved):

PR #207, opened 2022-07-31
fix: `no-reactive-literals` using `:matches(...)`

tivac:
One of the suggestions on #203 was to use `:matches(...)` around an array `.join()` which somehow got only... halfway implemented. The `.join()` was there, but it was missing the `:matches(...)` around it which could lead to some pretty bad false positives.

Added back the `:matches(...)` and added a few more valid test-cases to watch out for false-positives like this going forward.


ota-meshi (approved):
Thank you!

PR #208, opened 2022-07-31
Add `prefer-destructured-store-props` rule

tivac:
## :book: Rule Details

This rule reports on directly accessing properties of a store containing an object. These usages can instead be written as a reactive statement using destructuring to allow for more granular change-tracking and reduced redraws in the component.

An example of the improvements can be see in this [REPL](https://svelte.dev/repl/7de86fea94ff40c48abb82da534dfb89)

```svelte
<script>
  /* eslint svelte/prefer-destructured-store-props: "error" */
  $: ({ foo } = $store)
</script>

<!-- ✓ GOOD -->
{foo}

<!-- ✗ BAD -->
{$store.foo}
```

### DIFF
--- a/src/rules/prefer-destructured-store-props.ts
+++ b/src/rules/prefer-destructured-store-props.ts
@@ -1,0 +1,69 @@
+import type { TSESTree } from "@typescript-eslint/types"
+import { createRule } from "../utils"
+
+export default createRule("prefer-destructured-store-props", {
+  meta: {
+    docs: {
+      description: "",
+      category: "Best Practices",
+      recommended: false,
+    },
+    hasSuggestions: true,
+    schema: [],
+    messages: {
+      useDestructuring: `Destructure {{prop}} from store {{store}} for better change tracking & fewer redraws`,
+      fixUseDestructuring: `Using destructuring like $: ({ {{prop}} } = {{store}}); will run faster`,
+    },
+    type: "suggestion",
+  },
+  create(context) {
+    return {
+      // {$foo.bar + baz}
+      // should be
+      // $: ({ bar } = $foo);
+      // {bar + baz}
+      [`MemberExpression[object.name=/$/][property.type="Identifier"]`](
+        node: TSESTree.MemberExpression,
+      ) {
+        const store = (node.object as TSESTree.Identifier).name
+
+        console.log(node)
+
+        // Since the regex can't specify positioning of the "$", we need to check again
+        if (!store.startsWith("$")) {
+          return false
+        }
+
+        const prop = (node.property as TSESTree.Identifier).name
+
+        return context.report({
+          node,
+          messageId: "useDestructuring",
+          data: {
+            store,
+            prop,
+          },
+          suggest: [
+            {
+              messageId: "fixUseDestructuring",
+              data: {
+                store,
+                prop,
+              },
+
+              fix(fixer) {
+                return [
+                  fixer.insertTextBefore(
+                    node,
+                    `$: ({ ${prop} } = ${store});\n`,
+                  ),
+                  fixer.replaceText(node, prop),
+                ]
+              },
+            },
+          ],
+        })
+      },
+    }
+  },
+})
--- a/src/rules/prefer-destructured-store-props.ts
+++ b/src/rules/prefer-destructured-store-props.ts
@@ -3,7 +3,6 @@ import type { TSESTree } from "@typescript-eslint/types"
 import type { AST } from "svelte-eslint-parser"
 import { createRule } from "../utils"
 import { getStringIfConstant } from "../utils/ast-utils"
-import { returnStatement } from "@babel/types"
 
 type NodeRecord = { property: string; node: TSESTree.MemberExpression }
 
@@ -37,6 +36,16 @@ export default createRule("prefer-destructured-store-props", {
         script = node
       },
 
+      // Capture import names
+      [`ImportSpecifier, ImportDefaultSpecifier, ImportNamespaceSpecifier`](
+        node: TSESTree.ImportSpecifier,
+      ) {
+        const { name } = node.local
+
+        defined.add(name)
+      },
+
+      // Capture variable names
       [`VariableDeclarator[id.type="Identifier"]`](
         node: TSESTree.VariableDeclarator,
       ) {
--- a/src/rules/prefer-destructured-store-props.ts
+++ b/src/rules/prefer-destructured-store-props.ts
@@ -22,16 +22,10 @@ export default createRule("prefer-destructured-store-props", {
       // should be
       // $: ({ bar } = $foo);
       // {bar + baz}
-      [`MemberExpression[object.name=/$/][property.type="Identifier"]`](
+      [`MemberExpression[object.name=/^\\$/][property.type="Identifier"]`](
         node: TSESTree.MemberExpression,
       ) {
         const store = (node.object as TSESTree.Identifier).name
-
-        // Since the regex can't specify positioning of the "$", we need to check again
-        if (!store.startsWith("$")) {
-          return false
-        }
-
         const prop = (node.property as TSESTree.Identifier).name
 
         return context.report({
--- a/src/rules/prefer-destructured-store-props.ts
+++ b/src/rules/prefer-destructured-store-props.ts
@@ -1,6 +1,10 @@
+import type * as ESTree from "estree"
 import type { TSESTree } from "@typescript-eslint/types"
 import type { AST } from "svelte-eslint-parser"
 import { createRule } from "../utils"
+import { getStringIfConstant } from "../utils/ast-utils"
+
+type NodeRecord = { property: string; node: TSESTree.MemberExpression }
 
 export default createRule("prefer-destructured-store-props", {
   meta: {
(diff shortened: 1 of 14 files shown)


src/rules/prefer-destructured-store-props.ts
```diff
@@ -0,0 +1,67 @@
...
+    },
+    type: "suggestion",
+  },
+  create(context) {
+    return {
+      // {$foo.bar + baz}
+      // should be
+      // $: ({ bar } = $foo);
+      // {bar + baz}
+      [`MemberExpression[object.name=/$/][property.type="Identifier"]`](
```
ota-meshi, line 25: `/$/` matches the end, so I think it matches all strings. We use `store.startsWith("$")` to check for it, so I think we don't need this selector.

tivac: Yeah, that's been bugging me and I went back to esquery and I think it should actually be `/^\$/` which means we can remove the check entirely.

src/rules/prefer-destructured-store-props.ts
```diff
@@ -0,0 +1,61 @@
+import type { TSESTree } from "@typescript-eslint/types"
+import { createRule } from "../utils"
+
+export default createRule("prefer-destructured-store-props", {
+  meta: {
+    docs: {
+      description: "",
```
ota-meshi, line 7: Could you please enter a description?

tests/fixtures/rules/prefer-destructured-store-props/invalid/test01-errors.json
```diff
@@ -0,0 +1,38 @@
+[
+  {
+    "message": "Destructure bar from store $foo for better change tracking & fewer redraws",
+    "line": 2,
+    "column": 2,
+    "suggestions": [
+      {
+        "desc": "Using destructuring like $: ({ bar } = $foo); will run faster",
+        "messageId": "fixUseDestructuring",
+        "output": "<!-- prettier-ignore -->\n{$: ({ bar } = $foo);\nbar}\n{$foo[bar]}\n\n<!-- eslint-disable-next-line dot-notation -- tests -->\n{$foo.bar}\n"
```
ota-meshi, line 10: This suggestion looks strange to me.

```svelte
{$: ({ bar } = $foo);
bar}
{$foo[bar]}

<!-- eslint-disable-next-line dot-notation -- tests -->
{$foo.bar}
```

I think that at least reactive statements should be placed inside `<script>`.

ota-meshi (changes requested):

tests/fixtures/rules/prefer-destructured-store-props/invalid/test01-errors.json
tivac: I wasn't sure if it was worth writing all the logic to find/create a `<script>` for that. If you think it is, I can probably figure it out.

tests/fixtures/rules/prefer-destructured-store-props/invalid/test01-errors.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  {
+    "message": "Destructure bar from store $foo for better change tracking & fewer redraws",
+    "line": 7,
+    "column": 2,
+    "suggestions": [
+      {
+        "desc": "Using destructuring like $: ({ bar } = $foo); will run faster",
+        "messageId": "fixUseDestructuring",
+        "output": "<!-- prettier-ignore -->\n<script>\n\n$: ({ bar } = $foo);\n</script>\n\n{$foo.bar}\n{bar}\n\n<!-- eslint-disable-next-line dot-notation -- tests -->\n{$foo.bar}\n"
```
ota-meshi, line 22: I think the following suggestion is wrong.

```svelte
<!-- prettier-ignore -->
<script>
$: ({ bar } = $foo);
</script>

{$foo.bar}
{bar}

<!-- eslint-disable-next-line dot-notation -- tests -->
{$foo.bar}
```

I think it should be like this:

```svelte
<!-- prettier-ignore -->
<script>
$: ({ [bar]: barValue } = $foo);
</script>

{$foo.bar}
{barValue}

<!-- eslint-disable-next-line dot-notation -- tests -->
{$foo.bar}
```

It's hard to suggest this, so I think it's better not to suggest when using `[]`.

ota-meshi (changes requested):

tests/fixtures/rules/prefer-destructured-store-props/invalid/test01-errors.json
tivac: Fair!

src/rules/prefer-destructured-store-props.ts
```diff
@@ -0,0 +1,105 @@
...
+      [`Program:exit`]() {
+        reports.forEach(({ property, node }) => {
+          const store = (node.object as TSESTree.Identifier).name
+          // let prop: string | null = null
+
+          // if (node.property.type === "Literal") {
+          //   prop = node.property.value as string
+          // } else if (node.property.type === "Identifier") {
+          //   prop = node.property.name
+          // }
```
ota-meshi, line 63: This commenting out seems unnecessary. Could you remove it?

ota-meshi (changes requested):

src/rules/prefer-destructured-store-props.ts
```diff
@@ -0,0 +1,105 @@
...
+        const property =
+          node.property.type === "Identifier"
+            ? node.property.name
+            : getStringIfConstant(node.property as ESTree.Expression)
+
+        if (!property) {
+          return
+        }
+
+        reports.push({ property, node })
```
ota-meshi, line 51: Could you add the following test case? I tested it and noticed a false positive.

```svelte
<script>
  $: foo2 = $store.foo;
</script>

{foo2}
```

ota-meshi: Also, could you add the following test case? I think member access in reactive statements should probably be ignored.

```svelte
<script>
  import store from "./store.js";
	
  let foo, bar
  $: {
    foo = $store.foo;
    bar = $store.bar;
  }
</script>

<div>
  foo: {foo + " " + Date.now()}
</div>
<div>
  bar: {bar + " " + Date.now()}
</div>
```

Ideally it would be nice to use reactive variables, but I think that could be checked with another new rule.

tivac: > Could you add the following test case? I tested it and noticed a false positive.
> 
> ```svelte
> <script>
>   $: foo2 = $store.foo;
> </script>
> 
> {foo2}
> ```

I don't see the false positive here, can you elaborate?

https://svelte.dev/repl/1d043d248c0c4200b809ca84242301f1?version=3.49.0

tivac: > Also, could you add the following test case? I think member access in reactive statements should probably be ignored.

I think they should still trigger, but they shouldn't have suggestions. Added tracking of imported & defined names to not suggest if there would be a collision.

ota-meshi: I don't think you'll see any difference in render times between using destructuring and assigning values via member access.

https://svelte.dev/repl/897da31522c3410596f0dbc4916e5014?version=3.49.0

In my opinion, these member accesses should not be reported as the purpose of this rule is to reduce render times.
In addition to this rule, I think it makes sense to add a new rule that corrects member access to the destructuring writing style. The purpose of the new rules is to enforce stylistic conventions.
In my opinion, it makes sense to have separate rules because they serve different purposes.

What do you think?

ota-meshi: I just noticed that the REPL you provided in #205 seems to have incorrect results due to typo.

```diff
-	<code>direct = obj.a</code><br />{obj.a + " " + Date.now()}	
+	<code>direct = obj.a</code><br />{direct + " " + Date.now()}	
```

You wrote the following, but when I fix the typo there is no difference in the timestamps.

> Notice how often the timestamp is changing on the "obj.a" or "direct = obj.a" versions vs the "({ a } = obj)" version? Each change in that timestamp indicates another redraw for the element.

ota-meshi: Looking at this result, in my opinion, prefer-reactive-destructuring may also be better split rules.

ota-meshi: The purpose of `prefer-reactive-destructuring` might be seen as enforcing stylistic conventions rather than reducing rendering times.

ota-meshi: If we split the rules, it doesn't seem necessary to add a stylistic rule.
The `prefer-destructuring` rule and the new `svelte/prefer-reactive-destructuring` rule seem to report them.

```svelte
<script>
  /* eslint svelte/prefer-reactive-destructuring: error */
  $: foo2 = $store.foo;
</script>
```

```svelte
<script>
/* eslint prefer-destructuring: error */

import store from "./store.js";
  
let foo, bar
$: {
  foo = $store.foo;
  bar = $store.bar;
}
</script>
```

[prefer-destructuring DEMO]

[prefer-destructuring DEMO]: https://ota-meshi.github.io/eslint-plugin-svelte/playground/#eJxFjkEOgkAMAL/SNJwIsvcVfUkvAsWsgZZ0y4nwd1dM9DiTOcyOg46MEbs8WFr9ThJq4DwncViNJ7bLyNltG3yzJM8IbKYGdSAhScuq5pBdjWEyXYCwDSe2r0x4JQEgmdlhUm2gfxhJFWH/+GLgBtW3LnDGpfjLAkUeJF347WGDts2cMe7H8Qb6vEHv

src/rules/prefer-destructured-store-props.ts
```diff
@@ -0,0 +1,130 @@
...
+import { createRule } from "../utils"
+import { getStringIfConstant } from "../utils/ast-utils"
+
+type NodeRecord = { property: string; node: TSESTree.MemberExpression }
+
+export default createRule("prefer-destructured-store-props", {
+  meta: {
+    docs: {
+      description:
+        "Destructure values from object stores for better change tracking & fewer redraws",
```
marekvospel, line 13: The first letter should be lowercase to be consistent with existing rules. (see #218)

baseballyama:
@tivac What are the remaining tasks of the PR? I can do the remaining tasks.

tivac:
> @tivac What are the remaining tasks of the PR? I can do the remaining tasks if you need.

@baseballyama Same as #205, it's just a stylistic rule since there's no redraw-limiting reason to do it after Ota caught my typo. 🤦🏻‍♂️

tivac:
Sorry I've left these hanging for so long @ota-meshi, summer got busy and I haven't had much OSS time <3 Also realizing I based them off a false premise didn't help with motivation for getting them done, embarrassed I missed something like that.

ota-meshi:
I think the `svelte/prefer-destructured-store-props` rule probably needs to be changed to only report member accesses outside of `<script>`.
Other places where destructuring is preferred over member access can be reported by `svelte/prefer-reactive-destructuring` and `prefer-destructuring` as stylistic rules.

baseballyama:
> I think the svelte/prefer-destructured-store-props rule probably needs to be changed to only report member accesses outside of <script>.

It means, for example below code, should we report like "Please use `$ ({ foo } = $store)`"?

```sv
<script>
	import store from "./store.js";
</script>
<p>{$store.foo}</p>
```

ota-meshi:
> It means, for example below code, should we report like "Please use $ ({ foo } = $store)"?

That's right. I think users can reduce re-renders by doing so.

PR #209, opened 2022-08-02
Chore: Change it to use YAML instead of JSON as the fixture.

ota-meshi:
This PR changes it to use YAML as fixtures instead of JSON.

This change makes it easier to see the results of the suggestions.

image

~~Note: This PR must be merged after merging all currently open add rule PRs.~~
Changed to read json if only json exists so as not to block other PRs.


package.json
```diff
@@ -85,6 +85,7 @@
     "@types/eslint-scope": "^3.7.0",
     "@types/eslint-visitor-keys": "^1.0.0",
     "@types/estree": "^1.0.0",
+    "@types/js-yaml": "^4.0.5",
```
JounQin, line 88: Should we use more spec-compatible `yaml` package instead just like

https://github.com/ota-meshi/eslint-plugin-yml/blob/ef23f7d9a4744eaeef12bd219bedb46c999b09b1/package.json#L109

ota-meshi: I looked into it and js-yaml loads YAML faster. We don't use complex YAML, so compatibility is probably not an issue.

https://www.measurethat.net/Benchmarks/Show/20193/0/yaml-211-vs-js-yaml-410

JounQin: Thanks for letting me know about https://www.measurethat.net.

The benchmark does be significant, but it should not be a thing on current level of size project.

`js-yaml` is not added into `devDependencies`, can we guarantee where it comes from?

For `yaml`, it can come from `eslint-plugin-yml`.

ota-meshi: > The benchmark does be significant, but it should not be a thing on current level of size project.

I'm not too picky about it 😄, so I'll change it to `yaml`.

> js-yaml is not added into devDependencies, can we guarantee where it comes from?

I forgot to add it to devDeps.

JounQin (approved):

PR #212, opened 2022-08-05
Chore: fix doc site script

ota-meshi:
This PR fixes compilation issues on the docs site dev server when using the latest svelte-kit.

### DIFF
--- a/svelte.config.mjs
+++ b/svelte.config.mjs
@@ -19,6 +19,8 @@ fs.writeFileSync(
   JSON.stringify({ type: "module" }),
 )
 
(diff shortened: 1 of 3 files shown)


svelte.config.mjs
```diff
@@ -31,8 +33,8 @@ const config = {
     },
     adapter: ghpagesAdapter({
       // default options are shown
-      pages: "build",
-      assets: "build",
+      pages: outDir,
+      assets: outDir,
```
ota-meshi, line 37: This is changes the output destination of the documentation build files.
With this change, we can test static site by simply using the `npx http-server` command in `build` directory.

Issue #166, opened 2022-07-01
Use CI tools like `changesets` for changelog and releasing


JounQin:
https://github.com/changesets/changesets

Related bot https://github.com/apps/changeset-bot and action https://github.com/changesets/action

An example: https://github.com/import-js/eslint-import-resolver-typescript/pull/118/files

JounQin:
And also these projects seem lack of git hooks tools like [simple-git-hooks](https://github.com/toplenboren/simple-git-hooks) + [lint-staged](https://github.com/okonet/lint-staged)?

How do you think of them?

ota-meshi:
I agree to introduce those tools :+1:

JounQin:
@ota-meshi 

I'd like to work on this recently.

A few questions to discuss:

1. Do you want to release npm package manually or by CI automatically instead?
2. Do you have unified configs to be reused for tools like simple-git-hooks and lint-staged? Yours is always preferred here. If not can I use mine at https://github.com/1stG/configs/tree/master/packages Or they should just be used verbose.
3. Do we use https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional
4. Do we use https://github.com/remarkjs/remark-lint which can be integrated with ESLint via `eslint-mdx`

ota-meshi:
> Do you want to release npm package manually or by CI automatically instead?

I prefer manually release, but manually release is okay.

I don't know much about other tools as I haven't used them much. I think that it can be adjusted later if there is a problem after introducing it once.

JounQin:
OK, I'll install some GitHub Apps on this repo, it may require you to accept the installation.

JounQin:
@ota-meshi

I found I can not install any GitHub Apps due to lack of permission on a personal repository. 🤣

A GitHub organization is much more suitable for such cases.

Can you help to install them manually?

https://github.com/apps/changeset-bot/installations/new
https://github.com/apps/codesandbox-ci/installations/new
https://github.com/apps/lgtm-com/installations/new

ota-meshi:
I have installed 3 apps into this project and svelte-eslint-parser repo.
Secrets has also been set.

PR #216, opened 2022-08-11
feat(html-self-closing): add configuration presets

marekvospel:
see https://github.com/ota-meshi/eslint-plugin-svelte/pull/190#discussion_r931697359

### `html-self-closing`
added preset configuration option
```js
'svelte/html-self-closing': ['warn', {
  'void': 'always',
  // ...
}]
// or
'svelte/html-self-closing': ['warn', 'all']
```

current presets:
- none: all set to never
- html: html-compliant (always for voit and svelte* elements, others never)
- all: all on always

\* to be discussed whether svelte elements should or shouldn't be self closing in html-compliant preset.

## ~~Progress~~
~~this PR is still a draft, `html-self-closing` tests are still using the old configuration, no preset tests have been added and the documentation for `html-self-closing` rule is also outdates~~


marekvospel:
*sorry for the amount of commits this PR is going to have. I'm in an environment where I can't install packages, so I can't use type checking, code formatting etc.

marekvospel:
@ota-meshi @JounQin could you please generate errors.json & output.svelte for the preset tests? I won't be able to do so until saturday / sunday. I've checked `allow commits by maintainers`, so you should have permissions to make changes in my fork

JounQin:
I don't have time right now, I'd like to try when I'm free. 🍺

marekvospel:
The PR is ready! Once the invalid test outputs are generated and verified, it can be merged!

docs/rules/html-self-closing.md
```diff
@@ -54,7 +54,8 @@ You can choose either two styles for elements without content
 {
   "svelte/html-self-closing": [
     "error",
-    {
+    "all", // or "html" or "none" (preset)
+    {  // overrides
```
ota-meshi, line 58: I prefer `"all", "html", "none" or object` as the first option rather than splitting the options. I think the override option is confusing for users
What do you think about it?

marekvospel: good idea, this would also solve people having to migrate their configurations

ota-meshi:
We are about to introduce changeset and lgtm into this repository, but have not configured it yet. Ignore these messages for now.
#166

marekvospel:
I've updated the rule to take either a preset enum or config object, which makes this PR non-breaking

JounQin:
> I've updated the rule to take either a preset enum or config object, which makes this PR non-breaking

Right, so the PR description should be updated. 😁

marekvospel:
@ota-meshi @JounQin what do you think about the `html` preset? Should svelte special components like svelte:head be self closing?

src/rules/html-self-closing.ts
```diff
@@ -27,32 +27,65 @@ export default createRule("html-self-closing", {
...
-    const options = {
+    let options = {
       void: "always",
       normal: "always",
       component: "always",
       svelte: "always",
-      ...ctx.options?.[0],
+    }
+
+    if (typeof ctx.options?.[0] === "object") {
```
JounQin, line 63: What is `ctx.options?.[0] === null`?

src/rules/html-self-closing.ts
```diff
@@ -27,32 +27,65 @@ export default createRule("html-self-closing", {
...
+      options = {
+        ...options,
+        ...(ctx.options?.[0] ?? {}),
```
JounQin, line 66:
```suggestion
        ...(ctx.options?.[0]),
```

JounQin (changes requested):

JounQin:
> @ota-meshi @JounQin what do you think about the `html` preset? Should svelte special components like svelte:head be self closing?

According to Angular's HTML parser [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAhAEQPIGEAqAmgAoCiABABYwC2ANgHwA6UqAEvgLIAy5u3AQQDKQgLxMQUCAFoAVgGdyNQtNx0B8oROZRy5dqQGYdevak6l8AvmwEAlIZfEgArjABm0gBzaWp-fgAkvjcpAycAJ7kMMF0pKgA9EEhYX6m5pbWAHICFs4AJnDyYABOAJYADjBl0BLkkLAIMM6RfBBZMKRQ+L66+glshsZpqABGEPkRJunyAG5wdPBIlHAAhvkMiXMLSyvr02bbi3DLa-nkCQf6+WWzF9OJ45M6iRw8OiAANCAQVTVQ8mQoFWJRKEAA7sQQQhAShVnRwasIoDvqMSqswABrOAwISrGhwbhlKBwZDueHyOCo9FYnFCCoY4kAc2QMBKLipIDgNFGcHyhXy3FWUCZLlWTLgADEICUaKsYNURcgQKs3BAviBqPQAOqUMrweQMsBwIQw-W3fURZVgeQokDEyklGDEdFMuVkimchQADyEzLocAAii4IPAPXRKd8GSVHcrhaK6CCNRVyrBtWV8jBKMgvAAGKNgyna9EVZUpopwErzDUARxD8Bdv1hKvk0hJfL5GpKcDrZW7LvF7qQ5IjnMpNDKrPZY-9QfrpOHnu+MFWo3TmezSAATMv0WU6MzcBAaEOufIAKwalyU-Cr2EjyMgWYcwJQQqwISlSowARvoQwCIA3DSkAF8QKAA), `<svelte:head />` is html compliant, so yes.

JounQin:
> @ota-meshi @JounQin what do you think about the `html` preset? Should svelte special components like svelte:head be self closing?

According to Angular's HTML parser [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAhAEQPIGEAqAmgAoCiABABYwC2ANgHwA6UqAEvgLIAy5u3AQQDKQgLxMQUCAFoAVgGdyNQtNx0B8oROZRy5dqQGYdevak6l8AvmwEAlIZfEgArjABm0gBzaWp-fgAkvjcpAycAJ7kMMF0pKgA9EEhYX6m5pbWAHICFs4AJnDyYABOAJYADjBl0BLkkLAIMM6RfBBZMKRQ+L66+glshsZpqABGEPkRJunyAG5wdPBIlHAAhvkMiXMLSyvr02bbi3DLa-nkCQf6+WWzF9OJ45M6iRw8OiAANCAQVTVQ8mQoFWJRKEAA7sQQQhAShVnRwasIoDvqMSqswABrOAwISrGhwbhlKBwZDueHyOCo9FYnFCCoY4kAc2QMBKLipIDgNFGcHyhXy3FWUCZLlWTLgADEICUaKsYNURcgQKs3BAviBqPQAOqUMrweQMsBwIQw-W3fURZVgeQokDEyklGDEdFMuVkimchQADyEzLocAAii4IPAPXRKd8GSVHcrhaK6CCNRVyrBtWV8jBKMgvAAGKNgyna9EVZUpopwErzDUARxD8Bdv1hKvk0hJfL5GpKcDrZW7LvF7qQ5IjnMpNDKrPZY-9QfrpOHnu+MFWo3TmezSAATMv0WU6MzcBAaEOufIAKwalyU-Cr2EjyMgWYcwJQQqwISlSowARvoQwCIA3DSkAF8QKAA), `<svelte:head />` is html compliant, so yes.

src/rules/html-self-closing.ts
marekvospel: that is "object", however
```js
{
  test: 1,
  ...null,
}
// is
{
  test: 1,
}
```

JounQin: Then the reassignment is unnecessary, right? And it can be handled in the following `default` `case`, no need to handle it here?

src/rules/html-self-closing.ts
```diff
@@ -27,32 +27,65 @@ export default createRule("html-self-closing", {
...
+      }
+    } else {
+      switch (ctx.options?.[0] ?? "") {
```
JounQin, line 69:
```suggestion
      const option = ctx.options?.[0]
      switch (option) {
```

So that `option` can be reused in the default case as I described above.

JounQin (changes requested):

marekvospel:
I've realised the test outputs the errors, so I've copied it into the test output files, so the tests should pass now

marekvospel:
so many typos lmao 😂

JounQin (approved):
Thanks for your efforts!

ota-meshi (approved):
LGTM! Thank you!

PR #218, opened 2022-08-11
docs: uncapitalized first letters in rule desceiptions

marekvospel:
see https://github.com/ota-meshi/eslint-plugin-svelte/pull/186#discussion_r926289266

Rules added by @tivac have first letters in desceiption capitalized, which is inconsistent with the rest of the plugin.

Screenshot_20220811-211655_Firefox

This is also the case in #205 and #208.


ota-meshi (approved):
Thank you!

PR #225, opened 2022-08-14
feat: implement `stores-no-async`

baseballyama:
Related to: https://github.com/ota-meshi/eslint-plugin-svelte/issues/198 (`stores-no-async`)

This is the first ESLint rule I have written.
(So I'm sorry if I have a wrong understanding.)
I would be very grateful if you would review it.

### DIFF
--- a/.changeset/curvy-ants-admire.md
+++ b/.changeset/curvy-ants-admire.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-svelte": minor
+---
+
+Add svelte/stores-no-async rule by @baseballyama in #225
--- a/.changeset/curvy-ants-admire.md
+++ b/.changeset/curvy-ants-admire.md
@@ -2,4 +2,4 @@
 "eslint-plugin-svelte": minor
 ---
 
-Add svelte/stores-no-async rule by @baseballyama in #225
+Add svelte/stores-no-async rule
--- a/docs/rules/no-store-async.md
+++ b/docs/rules/no-store-async.md
@@ -1,0 +1,57 @@
+---
+pageClass: "rule-details"
+sidebarDepth: 0
+title: "svelte/no-store-async"
+description: "disallow using async/await inside svelte stores"
+since: "v3.1.0"
+---
+
+# svelte/no-store-async
+
+> disallow using async/await inside svelte stores
+
+- :gear: This rule is included in `"plugin:svelte/recommended"`.
+
+## :book: Rule Details
+
+This rule reports all uses of async/await inside svelte stores.
+Because it causes issues with the auto-unsubscribing features.
+
+<ESLintCodeBlock language="javascript">
+
+<!--eslint-skip-->
+
+```js
+/* eslint svelte/no-store-async: "error" */
+
+import { writable, readable, derived } from "svelte/store"
+
+/* ✓ GOOD */
+const w1 = writable(false, () => {})
+const r1 = readable(false, () => {})
+const d1 = derived(a1, ($a1) => {})
+
+/* ✗ BAD */
+const w2 = writable(false, async () => {})
+const r2 = readable(false, async () => {})
+const d2 = derived(a1, async ($a1) => {})
+```
+
+</ESLintCodeBlock>
+
+## :wrench: Options
+
+Nothing.
+
+## :books: Further Reading
+
+- [Svelte - Docs > 4. Prefix stores with $ to access their values / Store contract](https://svelte.dev/docs#component-format-script-4-prefix-stores-with-$-to-access-their-values-store-contract)
+
+## :rocket: Version
+
+This rule was introduced in eslint-plugin-svelte v3.1.0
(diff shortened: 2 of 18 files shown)


docs/rules/no-store-async.md
```diff
@@ -0,0 +1,57 @@
+---
+pageClass: "rule-details"
+sidebarDepth: 0
+title: "svelte/no-store-async"
+description: "disallow using async/await inside svelte stores"
+since: "v3.1.0"
```
baseballyama, line 6: We need to update here when we decide the release version of this rule.

src/rules/no-store-async.ts
```diff
@@ -0,0 +1,44 @@
...
+      unexpected: "Do not pass async functions to svelte stores.",
+    },
+    type: "problem",
+  },
+  create(context) {
+    return {
+      CallExpression(node: ESTree.CallExpression) {
+        if (node.callee.type !== "Identifier") return
+        const { name } = node.callee
+        if (name !== "writable" && name !== "readable" && name !== "derived")
```
ota-meshi, line 24: Could you change it to use eslint-utils? By using eslint-utils, we can also check for cases like:

```js
import * as stores from "svelte/store"
stores.writable()
```

https://eslint-utils.mysticatea.dev/api/scope-utils.html#referencetracker-class

I think the implementation of require-stores-init is helpful.
https://github.com/ota-meshi/eslint-plugin-svelte/blob/3362ad4053c554ad54f7c1e42193282f7d2fdaac/src/rules/require-stores-init.ts#L21-L35

src/rules/no-store-async.ts
```diff
@@ -0,0 +1,44 @@
...
+          return
+        const [, fn] = node.arguments
+        if (fn.type !== "ArrowFunctionExpression" || !fn.async) return
```
ota-meshi, line 27: I think we should check if the argument exists.

```suggestion
        if (!fn || fn.type !== "ArrowFunctionExpression" || !fn.async) return
```

ota-meshi, line 27: I think we should also check `"FunctionExpression"` ( `async function () {}` ).
(Though I don't think people will use it much.)

https://github.com/estree/estree/blob/master/es5.md#functionexpression
https://github.com/estree/estree/blob/master/es2017.md#function

docs/rules/no-store-async.md
ota-meshi: We don't need it now as it is automatically added by the release script.

```suggestion
```

.changeset/curvy-ants-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-svelte": minor
+---
+
+Add svelte/stores-no-async rule by @baseballyama in #225
```
ota-meshi, line 5: It's probably automatically linked. However, we've just introduced changesets, so I'm not sure if it works yet 😅 

```suggestion
Add svelte/stores-no-async rule
```

ota-meshi (changes requested):
Thank you for this PR!
I think it works mostly fine already, but I have some change requests to address some edge cases.

docs/rules/no-store-async.md
baseballyama: removed!
https://github.com/ota-meshi/eslint-plugin-svelte/pull/225/commits/c05ef7cc01b234f3627bd4f2209214169884d474

.changeset/curvy-ants-admire.md
baseballyama: fixed!
https://github.com/ota-meshi/eslint-plugin-svelte/pull/225/commits/715d04a1b1cc2cf013ae8462afa4f1f2a392c92b

src/rules/no-store-async.ts
baseballyama: updated!

https://github.com/ota-meshi/eslint-plugin-svelte/pull/225/commits/37d1de783bd7a6897a5f1b7377897576129ad485

I moved `extractStoreReferences` to utils folder but I'm not sure if this is your preferred way.

src/rules/no-store-async.ts
baseballyama: fixed! Sorry for that!

https://github.com/ota-meshi/eslint-plugin-svelte/pull/225/commits/99e0996c1866e285e6d3d4d92172f037415e3500

baseballyama: fixed!

https://github.com/ota-meshi/eslint-plugin-svelte/pull/225/commits/0b8b4ce6273f987ba0746df5c852cdf0a9b6fd5b

baseballyama:
@ota-meshi 

Thank you for your review!
I updated the PR according to your comments!
Could you please check it again? Thank you!

tivac (approved):

JounQin (approved):
I don't see anything strange. LGTM! 🍺

PR #228, opened 2022-08-15
chore: minor changes to the script

ota-meshi:


package.json
```diff
@@ -44,11 +44,10 @@
     "prebuild": "yarn clean",
     "prepare": "simple-git-hooks && yarn-deduplicate --strategy fewer || exit 0",
     "prerelease": "yarn clean && yarn build",
-    "pretest": "yarn build",
     "release": "changeset publish",
     "svelte-kit": "node --experimental-loader ./svelte-kit-import-hook.mjs node_modules/vite/bin/vite.js",
     "test": "yarn mocha \"tests/src/**/*.ts\" --reporter dot --timeout 60000",
-    "test:debug": "cross-env DEBUG=eslint-plugin-svelte* yarn test",
+    "test:debug": "env-cmd -e debug yarn test",
```
ota-meshi, line 50: Use env-cmd instead of cross-env.

package.json
```diff
@@ -44,11 +44,10 @@
     "prebuild": "yarn clean",
     "prepare": "simple-git-hooks && yarn-deduplicate --strategy fewer || exit 0",
     "prerelease": "yarn clean && yarn build",
-    "pretest": "yarn build",
```
ota-meshi, line 47: Remove `pretest `, as local testing takes time. Since it is also checked by CI, there should be no big problem.

.prettierignore
```diff
@@ -0,0 +1 @@
+.env-cmdrc
```
ota-meshi, line 1: We needed it because changes in `.env-cmdrc` caused errors in lint-staged.

JounQin: I'd like to fix this. 👍

JounQin (approved):

.prettierignore
JounQin: @ota-meshi 

OK, I just looked at `env-cmd`'s document, a [`.env-cmdrc`](https://github.com/toddbluhm/env-cmd#rc-file-usage file) is definitely a JSON file, the problem is our default prettier setting does not recognize `.*rc` files while my `@1stg/lint-staged` config knows that and will try to run `prettier --write` on `.env-cmdrc`.

Fix:

1. write `overrides` in `prettier.config.js`
2. use `@1stg/lint-staged` compatible `@1stg/prettier-config`
3. remove `@1stg/lint-staged` and adopt the verbose config

HDYT?

ota-meshi: I think `.env-cmdrc` is a JSON file as you say.
I didn't know that, but I think configuring overrides works fine.
https://prettier.io/docs/en/configuration.html#setting-the-parserdocsenoptionshtmlparser-option

I'm not sure what lint-staged configuration needs to be fixed. That's because I'm not familiar with lint-staged 😓 .

JounQin: > I'm not sure what lint-staged configuration needs to be fixed. That's because I'm not familiar with lint-staged 😓 .

`@1stg/lint-staged` is enabled and it recognizes `.env-cmdrc` as JSON file at https://github.com/1stG/configs/blob/master/packages/lint-staged/base.js#L18

> I didn't know that, but I think configuring overrides works fine.

Sure `overrides` will work as expected, see also

https://github.com/1stG/configs/blob/bda0754fa09108d8aef1a526acbf3cd44a5b33af/packages/prettier-config/base.js#L35-L41

---

Conclusion: so we are going to use `overrides` in `prettier.config.js` manually, right? I'd like to raise a PR for it.

PR #236, opened 2022-08-23
chore: change changesets config

ota-meshi:
See https://github.com/ota-meshi/eslint-plugin-svelte/pull/235#issuecomment-1223395556

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "ts": "node -r esbuild-register",
     "typecov": "type-coverage",
     "update": "yarn ts ./tools/update.ts && yarn format-for-gen-file",
-    "version": "env-cmd -e version yarn update && git add ."
+    "version": "env-cmd -e version yarn update && changeset version"
   },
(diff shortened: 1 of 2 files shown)


JounQin (approved):
Yeah, this is exactly what I mean!

package.json
```diff
@@ -51,7 +51,7 @@
     "ts": "node -r esbuild-register",
     "typecov": "type-coverage",
     "update": "yarn ts ./tools/update.ts && yarn format-for-gen-file",
-    "version": "env-cmd -e version yarn update && git add ."
+    "version": "env-cmd -e version yarn update && changeset version"
```
JounQin, line 54: Like `prerelease`,  we can extract `preversion`,  but it's not required, of course.

ota-meshi: We can't use preversion because the script needs to be run after package.json is updated.

JounQin: Then or `postversion`?

PR #248, opened 2022-09-03
chore: remove unused vetur settings

baseballyama:
This repo doesn't need to use vetur, so I removed it.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,8 +10,6 @@
   ],
   "editor.formatOnSave": true,
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "vetur.validation.script": false,
-  "vetur.validation.style": false,
   "editor.codeActionsOnSave": {
     "source.fixAll": true,
     "source.fixAll.stylelint": true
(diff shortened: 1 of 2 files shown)


baseballyama:
Hum... Why type error comes...

JounQin:
error @sveltejs/kit@1.0.0-next.464: The engine "node" is incompatible with this module. Expected version ">=16.14". Got "16.13.0"
error Found incompatible module.

baseballyama:
But still, I didn't get why suddenly type error comes from `cacheable-request`...?

https://github.com/ota-meshi/eslint-plugin-svelte/runs/8166580394?check_suite_focus=true#step:5:16

```
Error: node_modules/@types/cacheable-request/index.d.ts(26,42): error TS2709: Cannot use namespace 'ResponseLike' as a type.
```

JounQin:
Because we're not using lock files?

baseballyama:
`cacheable-request` is not updated recently. So I think type is not changed compared to other recent PRs even without lock files. This is weird...

JounQin:
I could also be affected by other global types.

baseballyama:
To avoid `cacheable-request` type error, I added `skipLibCheck` option to `tsconfig.json`.
I'm not sure this is the proper way, but now build and lint are succeeded.

ota-meshi (approved):
LTGM! Thank you!

PR #256, opened 2022-09-20
chore: update design

baseballyama:
related: https://github.com/ota-meshi/eslint-plugin-svelte/issues/253

I updated below points.

- Fix background color -> I stop to use gradation.
- change header behavior like [SvelteKit web site](https://kit.svelte.dev/) -> Now header has fixed position.
- Left side scroll bar's color. (Maybe SvelteKit web site7s color is better?) -> I followed Svelte docs color.

I'm not designer, so I don't have ability to create good design, but I think this PR is bit improved.
I'm open for your opinion.

image

### DIFF
--- a/docs-svelte-kit/src/app.css
+++ b/docs-svelte-kit/src/app.css
@@ -6,7 +6,7 @@
   --font-mono: "Fira Mono", monospace;
   --pure-white: #ffffff;
   --primary-color: #b9c6d2;
-  --secondary-color: #d0dde9;
+  --secondary-color: #676778;
   --tertiary-color: #edf0f8;
   --accent-color: #ff3e00;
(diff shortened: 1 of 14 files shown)


docs-svelte-kit/src/app.css
```diff
@@ -6,7 +6,7 @@
   --font-mono: "Fira Mono", monospace;
   --pure-white: #ffffff;
   --primary-color: #b9c6d2;
-  --secondary-color: #d0dde9;
+  --secondary-color: #676778;
```
baseballyama, line 9: `--secondary-color` is not used at all. So I updated this color according to Svelte site7s color.

ota-meshi (approved):
Great Job! Thank you!

Issue #258, opened 2022-09-22
Prettier error: `Parsing error: Adjacent JSX elements must be wrapped in an enclosing tag. Did you want a JSX fragment <>...</>?`


marekdedic:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.

### What version of ESLint are you using?

8.23.1

### What version of `eslint-plugin-svelte` are you using?

2.8.0

### What did you do?

<details>
<summary>Configuration</summary>

```js
module.exports = {
  parserOptions: {
    parser: "@typescript-eslint/parser",
    project: "./tsconfig.json",
    sourceType: "script",
    extraFileExtensions: [".svelte"],
  },
  env: {
    browser: true,
    node: false,
  },
  plugins: ["@typescript-eslint"],
  extends: [
    "eslint:recommended",
    "plugin:@typescript-eslint/recommended",
    "plugin:@typescript-eslint/recommended-requiring-type-checking",
    "plugin:svelte/recommended",
    "plugin:prettier/recommended",
    "plugin:svelte/prettier",
  ],
  overrides: [
    {
      files: ["*.svelte"],
      parser: "svelte-eslint-parser",
      parserOptions: {
        parser: "@typescript-eslint/parser",
      },
    },
  ],
};
```
</details>

```svelte
<h2>
  {$_("steps." + step + ".header")}
</h2>

<script lang="ts">
  import {_} from "svelte-i18n";

  export let step: string;
</script>

<style lang="scss">
  @import "@material/typography/mdc-typography";
</style>
```

### What did you expect to happen?

No prettier error

### What actually happened?

I got `Parsing error: Adjacent JSX elements must be wrapped in an enclosing tag. Did you want a JSX fragment <>...</>?` error from `eslint-plugin-prettier`

### Link to Minimal Reproducible Example

I can't get eslint-plugin-prettier running in your playground. For a (non-minimal) repro, see https://github.com/skaut/shared-drive-mover/pull/1077/commits/563b5a9e1df759736a319c85d8428baeede05ba6

JounQin:
Did you try `prettier-plugin-svelte`?

And I don't think that's a minimal reproduction, can you please try to provide another smaller one?

JounQin:
[...]

PR #259, opened 2022-09-23
More explicit docs for what the prettier config does

marekdedic:
Following up on the discussion in #258.


marekdedic:
Umm, I think the CI failure has nothing to do with my changes, let me know if I'm wrong...

ota-meshi:
I think it's caused by the svelte kit update. I will check it after tomorrow.

ota-meshi (approved):
Thank you!

PR #261, opened 2022-09-23
fix: false report in `settings.ignoreWarnings`

### LINKED ISSUE #257: Disable particular rules for template expressions
### Description

Hi,
I'd like to disable some rules for template expressions only - typically rules about typescript. Example:

```svelte
<button onClick={() => {clicked += 1}}>Increment</button> // HERE

<script lang="ts">
function fn() { // HERE
    return 3;
}
</script>
```

With my current configuration, I get the error `Missing return type on function` (`@typescript-eslint/explicit-function-return-type`) on the two lines marked by comments. This is problematic, because I can't use TS in template expressions and therefore I can't add a return type annotation to the first line of my snippet. However, I would still like to have the rule turned on for the second case, where it correctly reports that I forgot to add the return type annotation.

Currently, I found no way of doing this other than manually ignoring all the "false positives" or getting rid of any non-trivial template expressions - I would like to have a way to say to eslint that the rule `@typescript-eslint/explicit-function-return-type` should be turned off for all template expressions, however it should still check `<script />` blocks.

ota-meshi:
close #257

### DIFF
--- a/.changeset/chilly-olives-check.md
+++ b/.changeset/chilly-olives-check.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-svelte": patch
+---
+
+fix: false report in `settings.ignoreWarnings`
--- a/.github/workflows/NodeCI.yml
+++ b/.github/workflows/NodeCI.yml
@@ -50,7 +50,9 @@ jobs:
           rm -rf node_modules
         if: matrix.eslint != 8
       - name: Install Packages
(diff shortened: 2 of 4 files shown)


marekdedic:
Hi, I have tried this and can confirm it works :)

ota-meshi:
@marekdedic  Thank you for checking this PR.

PR #264, opened 2022-09-26
chore: improve svelte syntax in demo

ota-meshi:
before:
image

after: 
image


docs-svelte-kit/src/lib/eslint/scripts/syntax.js
```diff
@@ -0,0 +1,296 @@
+export const language = {
```
ota-meshi, line 1: Based on the html sample of https://microsoft.github.io/monaco-editor/monarch.html

PR #270, opened 2022-10-05
feat: add `svelte/prefer-destructured-store-props` rule

ota-meshi:
This PR is a follow up for #208.

Changed the rule to only warn about store property access in templates.

close #208


src/rules/prefer-destructured-store-props.ts
```diff
@@ -0,0 +1,340 @@
...
+  findAttribute,
+  isExpressionIdentifier,
+  findVariable,
+} from "../utils/ast-utils"
+
+type StoreMemberExpression = TSESTree.MemberExpression & {
+  object: TSESTree.Identifier & { name: string }
+}
+
+const keywords = new Set([
```
tivac, line 16: Would something like [`esutils`](https://www.npmjs.com/package/esutils#keyword) be useful here since it's less likely to drift out of sync?

tivac (approved):
Very thorough, wow. I don't see any red flags.

src/rules/prefer-destructured-store-props.ts
ota-meshi: Thank you for checking this PR. I think it would be great to replace with esutils as you suggested! I will change this PR.

tivac (approved):

JounQin (approved):

PR #289, opened 2022-10-31
feat: add `svelte/require-store-reactive-access` rule

### LINKED ISSUE #282: svelte/require-store-reactive-access
### Motivation

This is coming from https://github.com/sveltejs/svelte/issues/7984.

### Description

Disallow to render store itself. Need to use `$` prefix or `get` function.

This rule is fixable.

```svelte
<!-- before -->
<p>{storeValue}</p>
<!-- after -->
<p>{$storeValue}</p>

```

### Examples

```svelte
<script lang="ts">
  import { writable, get } from 'svelte/store';
  const storeValue = writable("hello");
</script>

<!-- ✓ GOOD -->
<p>{$storeValue}</p>
<p>{get(storeValue)}</p>

<!-- ✗ BAD -->
<p>{storeValue}</p>
```

### Additional comments

Svelte can define your own store. So we need to support it at some point. (But I think it's difficult. Is there any way to provide it to TypeScript users?)

https://svelte.dev/docs#component-format-script-4-prefix-stores-with-$-to-access-their-values-store-contract

ota-meshi:
This PR adds `svelte/require-store-reactive-access` rule that disallow to render store itself.

close #282


ota-meshi:
I realized that we also need to disallow non-rendering access to the raw store. I will changing this rule.

```svelte
<script lang="ts">
  import { _ } from 'svelte-i18n'
  $: message = $_('message') // OK
  $: message = _('message') // NG
</script>
```

ota-meshi:
I improved this rule.

PR #332, opened 2022-12-29
feat: add `svelte/Infinite-reactive-loop` rule

baseballyama:
close: https://github.com/ota-meshi/eslint-plugin-svelte/issues/299


tests/fixtures/rules/infinite-reactive-loop/valid/test08-input.svelte
```diff
@@ -0,0 +1,16 @@
...
+
+  const func = () => {
+    return {
+      then: (fun) => fun(),
+      catch: (fun) => fun(),
+    }
+  }
+
+  $: (() => {
+    // TODO: This is a false positive. we need to check if the `then` or `catch` is `Promise` object's function.
```
baseballyama, line 12: @ota-meshi 

I need your help!
I wanted to check that `then` or `catch` came from `Promise` type, but I couldn't find a way.
Could you please tell me?

ota-meshi: I think it's difficult to strictly check whether it's a Promise object or not. However, it's mostly okay to consider objects with `then` and `catch` methods that take functions as parameters to be Promises.
Other popular plugins also judge that way.

- https://github.com/typescript-eslint/typescript-eslint/blob/4ab9bd7f7c1812df4371d1fd3202969c1039e8a7/packages/eslint-plugin/src/rules/no-floating-promises.ts#L234
- https://github.com/eslint-community/eslint-plugin-promise/blob/6d046e6033d43d2c023966b33a7fabc1438899c2/rules/lib/is-promise.js#L14

It's clearly a false positive in this test case, but I think we can consider that people don't do it.

baseballyama:
@ota-meshi What do you think about the current specification and implementation? Do you think there are TODOs?

ota-meshi:
I misunderstood that this PR was still in the work in progress 😓. I will make time to check this PR. So please wait :pray:

src/rules/infinite-reactive-loop.ts
```diff
@@ -0,0 +1,439 @@
...
+      if (isReactiveVariableNode(context, node)) {
+        reactiveVariableNodes.push(node)
+      }
+    },
+    leaveNode() {
+      /* noop */
+    },
+  })
+  return reactiveVariableNodes
+}
```
ota-meshi, line 184:
```suggestion

/**
 * Get all reactive variable reference.
 */
function getAllReactiveVariableReferences(context: RuleContext) {
  const scopeManager = context.getSourceCode().scopeManager
  // Find the top-level (module or global) scope.
  // Any variable defined at the top-level (module scope or global scope) can be made reactive.
  const toplevelScope =
    scopeManager.globalScope?.childScopes.find(
      (scope) => scope.type === "module",
    ) || scopeManager.globalScope
  if (!toplevelScope) {
    return []
  }

  // Extracts all reactive references to variables defined in the top-level scope.
  const reactiveVariableNodes: TSESTree.Identifier[] = []
  for (const variable of toplevelScope.variables) {
    for (const reference of variable.references) {
      if (
        reference.identifier.type === "Identifier" &&
        !isFunctionCall(reference.identifier)
      ) {
        reactiveVariableNodes.push(reference.identifier)
      }
    }
  }
  return reactiveVariableNodes
}

/**
 * Get all tracked reactive variables.
 */
function getTrackedVariableNodes(
  context: RuleContext,
  ast: AST.SvelteReactiveStatement,
) {
  const reactiveVariableNodes: TSESTree.Identifier[] = []
  for (const identifier of getAllReactiveVariableReferences(context)) {
    if (
      // If the identifier is within the reactive statement range,
      // it is used within the reactive statement.
      ast.range[0] <= identifier.range[0] &&
      identifier.range[1] <= ast.range[1]
    ) {
      reactiveVariableNodes.push(identifier)
    }
  }
  return reactiveVariableNodes
}
```

src/rules/infinite-reactive-loop.ts
```diff
@@ -0,0 +1,439 @@
...
+    if (!resolved || resolved.name !== node.name) return false
+
+    return resolved.defs.some((def) => {
+      return (
+        (def as any).parent?.parent?.type === "SvelteScriptElement" &&
+        def.name.type === "Identifier" &&
+        def.name.name === node.name
+      )
+    })
+  })
```
ota-meshi, line 126: Using `getAllReactiveVariableReferences` as suggested in another comment:

```suggestion
  return getAllReactiveVariableReferences(context).includes(node)
```

src/rules/infinite-reactive-loop.ts
```diff
@@ -0,0 +1,439 @@
...
+            declaration = getDeclarationBody(child, functionCall.name)
+            if (declaration) break
+          }
+        }
+      }
+    }
+  }
+
+  return declaration
+}
```
ota-meshi, line 253: Note the addition of the `context` argument in the suggestion.

```suggestion
function getFunctionDeclarationNode(
  context: RuleContext,
  functionCall: TSESTree.Identifier,
): TSESTree.BlockStatement | TSESTree.Expression | null {
  const variable = findVariable(context, functionCall)
  if (!variable) {
    return null
  }
  for (const def of variable.defs) {
    if (def.type === "FunctionName") {
      if (def.node.type === "FunctionDeclaration") {
        return def.node.body
      }
    }
    if (def.type === "Variable") {
      if (
        def.node.init &&
        (def.node.init.type === "FunctionExpression" ||
          def.node.init.type === "ArrowFunctionExpression")
      ) {
        return def.node.init.body
      }
    }
  }
  return null
}
```

src/rules/infinite-reactive-loop.ts
```diff
@@ -0,0 +1,439 @@
...
+            isSameMicroTask = false
+          }
+        } else {
+          isSameMicroTask = false
+        }
+      }
+
+      // Promise.then() or Promise.catch() is called.
+      if (isPromiseThenOrCatchBody(node)) {
+        isSameMicroTask = true
```
ota-meshi, line 379: I think we can remove the duplicate code by keeping the `node` when it changes to `isSameMicroTask = false` in `enterNode()` so that when the same `node` comes, `isSameMicroTask = true`. What do you think?

ota-meshi (changes requested):
Looks pretty good to me, but I think using a scopeManager allows we to use already parsed variables and reduce we own traversal.

src/rules/infinite-reactive-loop.ts
baseballyama: Good catch. I removed the duplicate code.

https://github.com/ota-meshi/eslint-plugin-svelte/pull/332/commits/336dfa1949b39f6729b7ba3b2aaf63c0ce66b62e

ota-meshi (approved):
LGTM! Thank you!

PR #334, opened 2022-12-29
chore: Enable user scroll even if mouse is on a Monaco editor

baseballyama:
fix: https://github.com/ota-meshi/eslint-plugin-svelte/issues/331


ota-meshi (approved):
Wow! Thank you for finding that option!

PR #351, opened 2023-01-23
chore(deps): update sourcemap-codec dependency

SiNONiMiTY:
Update sourcemap-codec@1.4.8 dependency to @jridgewell/sourcemap-codec to remove package manager depreciation warnings during plugin installation.


ota-meshi (changes requested):
Thank you for this PR.

Could you also fix the source code?

https://github.com/ota-meshi/eslint-plugin-svelte/blob/f36be23bfbcaeca5977e9647409e64557426f022/src/shared/svelte-compile-warns/index.ts#L3-L4

ota-meshi (approved):
Thank you!

PR #354, opened 2023-02-01
Added the `require-event-dispatcher-types` rule

### LINKED ISSUE #346: New rule: `require-event-dispatcher-types`
### Motivation

Hi,
when using svelte with TypeScript, it is possible to add types to events by typing the `createEventDispatcher` function so:

```svelte
import { createEventDispatcher } from "svelte";

const dispatch = createEventDispatcher<{one: never, two: number}>();

dispatch("one"); // OK
dispatch("two", 3); // OK
dispatch("one", "Hello world"); // Fails
dispatch("two"); // Fails
dispatch("two", "Hello world"); // Fails
```

I propose a rule that would require all calls to `createEventDispatcher` to include the templated type.

### Description

The rule would trigger on any call of `createEventDispatcher` witchou the templated type

### Examples

```svelte
import { createEventDispatcher } from "svelte";

<!-- ✓ GOOD -->
const dispatch = createEventDispatcher<{one: never, two: number}>();

const dispatch = createEventDispatcher<any>();

const dispatch = createEventDispatcher<{}>();

<!-- ✗ BAD -->
const dispatch = createEventDispatcher();
```

### Additional comments

_No response_

marekdedic:
Closes #346

### DIFF
--- a/src/rules/require-event-dispatcher-types.ts
+++ b/src/rules/require-event-dispatcher-types.ts
@@ -1,4 +1,5 @@
 import { createRule } from "../utils"
+import { getLangValue } from "../utils/ast-utils"
 import { extractCreateEventDispatcherReferences } from "./reference-helpers/svelte-createEventDispatcher"
 
 export default createRule("require-event-dispatcher-types", {
@@ -15,8 +16,18 @@ export default createRule("require-event-dispatcher-types", {
     type: "suggestion",
   },
   create(context) {
+    let isTs = false
     return {
-      Program() {
+      SvelteScriptElement(node) {
+        const lang = getLangValue(node)?.toLowerCase()
+        if (lang === "ts" || lang === "typescript") {
+          isTs = true
+        }
+      },
+      "Program:exit"() {
+        if (!isTs) {
+          return
+        }
         for (const node of extractCreateEventDispatcherReferences(context)) {
           if (node.typeParameters === undefined) {
             context.report({ node, messageId: "missingTypeParameter" })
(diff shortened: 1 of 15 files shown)


src/rules/require-event-dispatcher-types.ts
```diff
@@ -0,0 +1,28 @@
...
+    },
+    schema: [],
+    messages: {
+      missingTypeParameter: `Type parameters missing for the createEventDispatcher function call.`,
+    },
+    type: "suggestion",
+  },
+  create(context) {
+    return {
+      Program() {
```
ota-meshi, line 19: I think we need to check if `lang="ts"` is specified before we check it.

```suggestion
    let isTs = false
    return {
      SvelteScriptElement(node) {
        const lang = getLangValue(node)?.toLowerCase()
        if (lang === "ts" || lang === "typescript") {
          isTs = true
        }
      },
      "Program:exit"() {
        if (isTs) return
```

tests/fixtures/rules/require-event-dispatcher-types/invalid/no-types01-input.svelte
```diff
@@ -0,0 +1,5 @@
+<script lang="ts">
```
ota-meshi, line 1: Could you add another test case without `lang="ts"`?

ota-meshi (changes requested):
Thank you for this PR! I made some comments.

marekdedic:
Thanks for the review :) It's a good catch that the rule should probably ignore non-ts code outright.

ota-meshi (approved):
LGTM! Thank you!

PR #365, opened 2023-02-07
feat: Added the `experimental-require-strict-events` rule

### LINKED ISSUE #345: New rule: `require-strict-events`
### Motivation

Hi,
when using TypeScript with svelte, component events aren't automatically typed - e. g. you can have an `on:nonexistent={() => {}}` handle for a non-existent event. However, svelte does check this, if the opening script tag of the *child* component has the `strictEvents` attribute. (`<script lang="ts" strictEvents>`).

I propose to add a rule that would enforce this.

### Description

The rule should enforce that components have the `strictEvents` attribute present in their opening svelte tag.

I propose several an option to choose for which components the attribute should be enforced:
1. Require the attribute on all components. This makes sense since for a component that does not dispatch any event, the attribute still makes svelte check that no event listeners are present. I propose that this should be the default.
2. Only check components that actually dispatch an event - this would probably be more complicated to implement

Additional note: For some components that have event mixins, the `strictEvents` attribute does not work and you have to declare the `$$Events` interface. In that case, I think the rule shouldn't fire - so effectively, it would require *either* the `strictEvents` attribute or a manually declared interface `$$Events` for the component.

### Examples

```svelte
<!-- ✓ GOOD -->
<script lang="ts" strictEvents>
</script>

<script lang="ts">
  interface $$Events {}
</script>

<!-- ✗ BAD -->
<script lang="ts">
</script>
```

### Additional comments

The relevant [RFC](https://github.com/dummdidumm/rfcs/blob/ts-typedefs-within-svelte-components/text/ts-typing-props-slots-events.md#typing-events).

marekdedic:
Closes #345


marekdedic:
Hi @ota-meshi, there are some lint issues but I think they're unrelated to the PR...

ota-meshi:
> there are some lint issues but I think they're unrelated to the PR...

I think merging or rebasing the current main branch will probably solve it. Could you try it?

marekdedic:
> I think merging or rebasing the current main branch will probably solve it. Could you try it?

No luck :(

tools/render-rules.ts
```diff
@@ -7,10 +7,11 @@ const categories = [
   "Best Practices",
   "Stylistic Issues",
   "Extension Rules",
+  "Experimental",
   "System",
 ] as const
 
-const descriptions: Record<(typeof categories)[number], string> = {
+const descriptions: Record<typeof categories[number], string> = {
```
ota-meshi, line 14: I think you need to revert it.

marekdedic: Hmm, how did that even happen?

src/rules/experimental-require-strict-events.ts
```diff
@@ -0,0 +1,45 @@
...
+      },
+      TSInterfaceDeclaration(node) {
+        if (node.id.name === "$$Events") {
+          hasInterface = true
+        }
+      },
+      "Program:exit"() {
+        if (isTs && !hasAttribute && !hasInterface) {
+          context.report({
+            loc: {
```
ota-meshi, line 35: By reporting to the position of the opening tag of the `<script>`, I think that the report will be more understandable for the user.

src/rules/experimental-require-strict-events.ts
```diff
@@ -0,0 +1,45 @@
...
+        isTs = lang === "ts" || lang === "typescript"
+        hasAttribute = findAttribute(node, "strictEvents") !== null
+      },
+      TSInterfaceDeclaration(node) {
+        if (node.id.name === "$$Events") {
+          hasInterface = true
+        }
+      },
+      "Program:exit"() {
+        if (isTs && !hasAttribute && !hasInterface) {
```
ota-meshi, line 33: Do you think the `strictEvents` attribute should be given even if `createEventDispatcher` or `on:click` are not used?

ota-meshi (changes requested):
Thank you for this PR!
I made two comments. Could you please check them out?

src/rules/experimental-require-strict-events.ts
marekdedic: Yes, I do - the `strictEvents` attributes makes all "callers" of the component (i.e. components that contain it) have their `on:something` handlers checked as well - so even if a component has no events, the `strictEvents` attribute brings a benefit in that the callers will error if they try to add a handler to a non-existent event.

ota-meshi (approved):
LGTM! Thank you!

marekdedic:
Great, I'll get #368 ready :)

PR #375, opened 2023-02-12
Allow to specify TS target using parserOptions

ptrxyz:
Suggestion for https://github.com/ota-meshi/eslint-plugin-svelte/issues/374

### DIFF
--- a/src/shared/svelte-compile-warns/transform/typescript.ts
+++ b/src/shared/svelte-compile-warns/transform/typescript.ts
@@ -29,7 +29,7 @@ export function transform(
     const output = ts.transpileModule(code, {
       reportDiagnostics: false,
       compilerOptions: {
-        target: ts.ScriptTarget.ESNext,
+        target: context.parserOptions.target || ts.ScriptTarget.ESNext,
         importsNotUsedAsValues: ts.ImportsNotUsedAsValues.Preserve,
         sourceMap: true,
       },
--- a/src/shared/svelte-compile-warns/transform/typescript.ts
+++ b/src/shared/svelte-compile-warns/transform/typescript.ts
@@ -29,7 +29,7 @@ export function transform(
     const output = ts.transpileModule(code, {
       reportDiagnostics: false,
       compilerOptions: {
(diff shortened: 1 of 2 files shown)


src/shared/svelte-compile-warns/transform/typescript.ts
```diff
@@ -29,7 +29,7 @@ export function transform(
     const output = ts.transpileModule(code, {
       reportDiagnostics: false,
       compilerOptions: {
-        target: ts.ScriptTarget.ESNext,
+        target: context.parserServices.program.getCompilerOptions().target || ts.ScriptTarget.ESNext,
```
ota-meshi, line 32: Thank you for this PR.
Just to be sure, could you please change the following to use optional chaining?

```suggestion
        target: context.parserServices.program?.getCompilerOptions()?.target || ts.ScriptTarget.ESNext,
```

ota-meshi (changes requested):

src/shared/svelte-compile-warns/transform/typescript.ts
ptrxyz: Yep, done.

src/shared/svelte-compile-warns/transform/typescript.ts
```diff
@@ -29,7 +29,7 @@ export function transform(
     const output = ts.transpileModule(code, {
       reportDiagnostics: false,
       compilerOptions: {
-        target: ts.ScriptTarget.ESNext,
+        target: context.parserServices.program?.getCompilerOptions()?.target || ts.ScriptTarget.ESNext,
```
ota-meshi, line 32: It doesn't seem to work for our test cases. I think we probably need to add an option like this:

```suggestion
        target: context.parserServices.program?.getCompilerOptions()?.target || ts.ScriptTarget.ESNext,
        module: ts.ModuleKind.ESNext,
```

Also, Could you format it using `yarn lint-fix`?

ota-meshi (changes requested):

ota-meshi (changes requested):
Could you format it using yarn lint-fix?

ota-meshi (approved):
LGTM!

PR #389, opened 2023-03-01
feat: Added the `block-lang` rule

### LINKED ISSUE #380: New rule: `block-lang`
### Motivation

This rule disallows the use of languages other than those available in your application for the lang attribute of block elements.

On top of that, this would complement #345, #346 and #347

### Description

This rule would report all components not in the given language.

Rule configuration example:

```
{
  script: ["ts", "typescript", null],
  style: [null]
}
```

So you can set for both `<script>` and `<style>` blocks a list of allowed languages (where `null` means no language set). Default value would be `{script: null, style: null}` - using a shorthand omitting the array would be allowed like this

### Examples

with `{script: ["ts", "typescript"]}`:

```svelte
<!-- ✓ GOOD -->
<script lang="ts"></script>

<script lang="typescript"></script>

<!-- ✗ BAD -->
<script></script>

<script lang="js"></script>

<script lang="javascripts"></script>
```

with `{script: ["js", "null"]}`:

```svelte
<!-- ✓ GOOD -->
<script lang="js"></script>

<script></script>

<!-- ✗ BAD -->

<script lang="javascript"></script>

<script lang="typescript"></script>

<script lang="ts"></script>
```

### Additional comments

_No response_

marekdedic:
Closes #380


ota-meshi (commented):
Thank you for this PR!
I haven't had time to check this yet, but let me comment on the rule name.

In `*.vue`, `<script>` and `<style>` tags are called `block`.
https://vuejs.org/api/sfc-spec.html#overview
But it's probably not called `block` in `*.svelte`.

So I think it probably makes more sense to rename the rule. 
What do you think about naming it `lang-attribute` for example?

marekdedic:
Hi,
you'll get to it when you'll get to it, no pressure.

Actually, they're referred to as blocks in the svelte docs:
- https://svelte.dev/docs#component-format-script
- https://svelte.dev/docs#component-format-style

But I'm open to renaming it :)

ota-meshi:
Oh! Thank you! I didn't know that.
Thanks to the information you gave me, I had second thoughts that the name `block-lang` made the most sense.

ota-meshi (approved):
LGTM! Thank you very much!

PR #394, opened 2023-03-06
Fixed `block-lang` false positives for non-svelte files

### LINKED ISSUE #393: `block-lang` triggers on non-svelte files
Hi
`block-lang` reports plain `.ts` files as well, no just svelte files - kind of a false positive.

I'll try to fix it, but I am not sure how to find out whether a file is a svelte file - @ota-meshi, do you just check for file extension? I can't check for `<script>` or `<style>` blocks as they theoretically may not be present...

marekdedic:
Closes #393


marekdedic:
Hmm, I can't seem to get this to work :/ I have looked at the tests for `require-store-reactive-access` but if I move the `non-svelte` dir outside `valid` it does not fail when it should (commented out the fix - does this maybe mean those tests don't run for `require-store-reactive-acccess` as well?) and also `@typescript-eslint/no-unnecessary-condition` but that didn't work for me as well... :( I can drop the TS test if it's too complicated but maybe I am just missing something simple?

ota-meshi:
I made a change to support test fixtures in `*.ts` files. #395
Could you merge it into your branch?

ota-meshi:
> Could you merge it into your branch?

I just realized I can merge it 😅

ota-meshi (approved):
LGTM! Thank you!

Issue #401, opened 2023-03-09
Rule proposal: no global classes/ids


marekdedic:
### Motivation

Hi,
as I am converting an existing codebase to svelte, I encountered some element classes/IDs that no longer do anything have been left over. They were previously used either for stuff like `document.getElementById()` or just global CSS that I over time replaced by component styles.

I thought this could be useful to automate by an eslint rule:

### Description

The rule would disallow any element class or id attributes unless they are used in the component CSS.

### Examples

```svelte
<!-- ✓ GOOD -->
<div />

<!-- ✓ GOOD -->
<div class="my class" />

<style>
.myClass {}
</style>

<!-- ✗ BAD -->
<div class="myClass" />
```

### Additional comments

Does this make sense to you as a rule? I
Also, I'm not sure about the name. Also also, there should probably be a whitelist :).

marekdedic:
Ok, let's start with classes first - I still think IDs make sense as well, but you're right, there are many elements where IDs are use for different things than just styles, so it'd be much more complicated

PR #402, opened 2023-03-11
feat: Added the `no-unused-class-name` rule

marekdedic:


marekdedic:
Well, this is going to be interesting... :D 
@ota-meshi if I understand it correctly, `svelte-eslint-parser` doesn't actually parse the styles - `SvelteStyleElement.children` is just a `SvelteText` with the raw CSS - this isn't gonna be enough for this rule. Any plans to parse the CSS the same way as JS in `svelte-eslint-parser`? (And yes, it's going to need to support languages like SCSS and preprocessors as well - like it supports TS).

ota-meshi:
I currently have no plans to allow `svelte-eslint-parser` to parse `<style>`. Converting many CSS flavors into an AST that ESLint understands is difficult.

To implement the `no-unused-class-name` rule, I think we need to extract the CSS of the `<style>` tag and parse it with [postcss]. Next, I think it is necessary to extract the selector from the `postcss` AST and parse it with [postcss-selector-parser].

[postcss]: https://www.npmjs.com/package/postcss
[postcss-selector-parser]: https://www.npmjs.com/package/postcss-selector-parser

If you want to support CSS preprocessors, please refer to the code around the following link.
https://github.com/ota-meshi/eslint-plugin-svelte/blob/09b01ad1ce208c6dbbf8357914d3aadb71c57a4e/src/shared/svelte-compile-warns/index.ts#L24

baseballyama:
Why do we need this rule?
Already Svelte compiler itself has this feature.

ota-meshi:
The purpose of this rule is to detect classes specified on elements that were not used in CSS. #401.
It does not detect unused selectors (classes) in CSS.
I think this rule can be useful if your project doesn't use classes outside of `*.svelte` CSS.

baseballyama:
Got it!

marekdedic:
Waiting for sveltejs/svelte-eslint-parser#318

marekdedic:
Obsoleted by #489

PR #407, opened 2023-03-17
`all` config

### LINKED ISSUE #369: Export an all config
### Description

Export an all config that contains the default options for all of the rules (except for deprecated ones).

```js
module.exports = {
  extends: ['plugin:svelte/all'],
}
```

DetachHead:
fixes #369

### DIFF
--- a/.changeset/all-config.md
+++ b/.changeset/all-config.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-svelte": minor
+---
+
+add `all` config which enables every rule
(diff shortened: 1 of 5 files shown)


ota-meshi (changes requested):
Please document the caveats as mentioned in the issue. I think we also need to export from index.ts. Also we need to resolve the CI error.

DetachHead:
added the documentation and the export. as for the CI error, i have no idea what's going on. as far as i can tell it's unrelated because locally when i run the tests after reverting my changes the errors still occur

ota-meshi:
Thank you for the PR.  As you said, the remaining CI errors appear unrelated to the changes in the PR.  I will look into the cause later.

ota-meshi (approved):
LGTM! Thank you!

PR #414, opened 2023-03-23
fix syntax error in experimental-require-slot-types example

DetachHead:
this example was incorrectly in a javascript `script` block when it uses typescript syntax:

image


ota-meshi (approved):
Thank you!

PR #440, opened 2023-04-22
feat: add `svelte/no-reactive-reassign` rule

ota-meshi:
This PR adds `svelte/no-reactive-reassign` rule.

`svelte/no-reactive-reassign` rule that disallow reassigning reactive values.

e.g. 

```svelte
<script>
  /* eslint svelte/no-reactive-reassign: "error" */
  let value = 0
  $: reactiveValue = value * 2
  function handleClick() {
    /* ✓ GOOD */
    value++
    /* ✗ BAD */
    reactiveValue = value * 3
    reactiveValue++
  }
</script>
<!-- ✓ GOOD -->
<input type="number" bind:value />
<!-- ✗ BAD -->
<input type="number" bind:value={reactiveValue} />
```

### DIFF
--- a/.changeset/tidy-pugs-draw.md
+++ b/.changeset/tidy-pugs-draw.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-svelte": minor
+---
+
+feat: add `svelte/no-reactive-reassign` rule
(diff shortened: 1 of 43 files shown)


baseballyama (approved):
I found that the below doesn't detect the ESLint error, but I think it's too difficult.😅

```
<script>
  /* eslint svelte/no-reactive-reassign: "error" */
  let object = { value: 0 }
  $: reactiveValue = { value: object.value * 2, foo: { bar: object.value * 2 } }
  function handleClick() {
    const a = reactiveValue;
    // It should be error, but it's too difficult to check...
    delete a["foo"];
  }
</script>
```

ota-meshi:
Oh, you're right. However, it is difficult to report it accurately as you say, so I would like to consider it if there is a request from the user in the future.

PR #456, opened 2023-04-29
chore: add Node 20 in CI

baseballyama:
Svelte 4 branch also added CI with Node 20. So I added it to this repo also.


ota-meshi (approved):
Thank you!

PR #473, opened 2023-05-09
feat: add `svelte/require-each-key` rule

### LINKED ISSUE #459: New Rule: `svelte/require-each-key`
### Motivation

This is exactly the same as Vue rule.
https://eslint.vuejs.org/rules/require-v-for-key.html

### Description

Require `key` with `{#each}` syntax

### Examples

```svelte
<!-- ✓ GOOD -->
{#each items as item (item.id)}
  <li>{item.name} x {item.qty}</li>
{/each}

<!-- ✗ BAD -->
{#each items as item}
  <li>{item.name} x {item.qty}</li>
{/each}
```

### Additional comments

_No response_

ota-meshi:
close #459

### DIFF
--- a/src/rules/require-each-key.ts
+++ b/src/rules/require-each-key.ts
@@ -1,0 +1,27 @@
+import type { AST } from "svelte-eslint-parser"
+import { createRule } from "../utils"
+
+export default createRule("require-each-key", {
+  meta: {
+    docs: {
+      description: "require keyed `{#each}` block",
+      category: "Best Practices",
+      recommended: false,
+    },
+    schema: [],
+    messages: { expectedKey: "Each block should have a key" },
+    type: "suggestion",
+  },
+  create(context) {
+    return {
+      SvelteEachBlock(node: AST.SvelteEachBlock) {
+        if (node.key == null) {
+          context.report({
+            node,
+            messageId: "expectedKey",
+          })
+        }
+      },
+    }
+  },
+})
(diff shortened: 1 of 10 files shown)


ota-meshi:
As you said, I think it would be useful for a rule to check if a key contains an iteration variable :+1:
However, I think it would be better to implement the check as a separate rule.
Because even if we don't want the key to be required, we want the rule to check if the key contains an iteration variable.

src/rules/require-each-key.ts
```diff
@@ -0,0 +1,27 @@
+import type { AST } from "svelte-eslint-parser"
+import { createRule } from "../utils"
+
+export default createRule("require-each-key", {
+  meta: {
+    docs: {
+      description: "require keyed `{#each}` block",
+      category: "Best Practices",
+      recommended: false,
```
baseballyama, line 9: I think we can toggle to true in major version up?

baseballyama (approved):
I wrote one comment but LGTM!

src/rules/require-each-key.ts
ota-meshi: Hmm. I personally like to enforce the key, but `vue/require-v-for-key` is unpopular even in eslint-plugin-vue, so I think it's better not to include it in the preset. (However, the Vue style guide recommends using keys.)
I think it will be easier to introduce eslint if the presets are not too strict.

baseballyama: Okay!

PR #475, opened 2023-05-10
feat: add `svelte/valid-each-key` rule

ota-meshi:
See https://github.com/sveltejs/eslint-plugin-svelte/pull/473#pullrequestreview-1418940020


baseballyama (approved):
Why so fast?🚀

Issue #448, opened 2023-04-26
New Rule: `svelte/no-getContext-in-template`


baseballyama:
### Motivation

This is came from https://github.com/sveltejs/svelte/pull/7864

### Description

According to the docs, `getContext` needs to call during component initialization.
Svelte team tried to implement this in the compiler, but this is a bit difficult, so I thought that we will implement this as ESLint rule.

https://svelte.dev/docs#run-time-svelte-getcontext

> Retrieves the context that belongs to the closest parent component with the specified key. Must be called during component initialisation.

### Examples

```svelte
<script>
  import { getContext } from 'svelte';

  let test = getContext('test');
</script>



<!-- ✓ GOOD -->
<a href={test}>xxx</a>

<!-- ✗ BAD -->
<a href={getContext('test')}>xxx</a>
```

ota-meshi:
I have read the documentation. It seems that `getContext` isn't the only thing that has to be called during component initialization. So other functions such as `setContext`, `hasContext` should also be checked. Also, it seems necessary to warn when used inside `$:`, not just in templates.
Additionally, I think the rule name should be something like `valid-context-access`.

PR #480, opened 2023-05-14
feat: Add `svelte/valid-context-access` rule

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/448


src/rules/valid-context-access.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /** Let's lint! */
+    function doLint(
+      visitedCallExpressions: TSESTree.CallExpression[],
+      contextCallExpression: TSESTree.CallExpression,
+      currentNode: TSESTree.CallExpression,
+    ) {
+      let { parent } = currentNode
+      if (parent?.type !== "ExpressionStatement") {
```
ota-meshi, line 63: I suspect an if statement like the following will not work:

```js
if (hasContext("answer")) {
}
```
Could you add this test case?

src/rules/valid-context-access.ts
```diff
@@ -0,0 +1,113 @@
...
+      recommended: false,
+    },
+    schema: [],
+    messages: {
+      unexpected:
+        "Do not call {{function}} except during component initialization.",
+    },
+    type: "problem",
+  },
+  create(context) {
```
ota-meshi, line 21: It probably doesn't work well other then `*.svelte` files, so I think the following guard is needed.

```suggestion
  create(context) {
    if (!context.parserServices.isSvelte) {
      return {}
    }
```

ota-meshi (changes requested):
Thank you for this PR!
I haven't checked all yet, but I have some comments.

src/rules/valid-context-access.ts
baseballyama: I added a basic test for if statement. I think it works fine.
Is this same understanding with you?

https://github.com/sveltejs/eslint-plugin-svelte/pull/480/commits/a86eac8312032e1008f84c67b2a9b59590d71bc6

ota-meshi: Sorry. I should have explained more.
I would like to add a test for if statements are used at the top level.

```svelte
<script>
  import { hasContext } from "svelte"

  if (hasContext("answer")) {
    console.log("The answer exist")
  }
</script>
```

src/rules/valid-context-access.ts
baseballyama: I need to add more test but even JS file can use `onMount` so I think we need to check JS files also.

```js
import { setContext, onMount } from "svelte"

const something = () => {
  setContext("answer", 42)
}

onMount(() => something())

```

But at the same time, I found this pattern.
It should be work but now ESLint error occurs.
So I need to handle this.

```js
import { setContext } from "svelte"

const something = () => {
  setContext("answer", 42)
}

const wrapper = (fn) => {
  fn()
}

wrapper(() => something())
```

In addition, I forgot to handle async/await.😇

```js
const something = async () => {
  await Promise.resolve()
  setContext("answer", 42)
}

something()
```

src/rules/valid-context-access.ts
baseballyama: Oh OK! I will add a test.

baseballyama:
I realized that it’s impossible to check async/await things perfectly.
For example below case, it will be an error but I think it isn't easy to detect it by ESLint.

So my opinion is that we will mention this limitation on the docs and we don't check async/await things.
Or we don't implement this rule.
There may not be much value in the rule that includes this false negative.
What do you think @ota-meshi ?

```svelte
<script>
import { setContext } from "svelte"
import someAsyncProcess from './outside'

someAsyncProcess(() => {
  setContext("answer", 42)
});
```

```js
// outside.js
export async function someAsyncProcess(fn) {
  await Promise.resolve();
  fn();
}
```

ota-meshi:
Hmm. You're right, there are false negatives, but I don't think users will use setContext that way, so I don't think it matters much if it's not reported.

I think code like the following is a common mistake, so I think there's value in having it reported by a rule.

```js
const something = async () => {
  await Promise.resolve()
  setContext("answer", 42)
}
```

src/rules/infinite-reactive-loop.ts
```diff
@@ -1,57 +1,11 @@
...
-  })
-  return Array.from(a).map(({ node, path }) => {
-    return {
-      node: node as TSESTree.CallExpression,
-      name: path[path.length - 1],
-    }
-  })
-}
+import { extractSvelteLifeCycleReferences } from "./reference-helpers/svelte-lifecycle"
+import { extractTaskReferences } from "./reference-helpers/microtask"
```
baseballyama, line 8: I moved few functions to utils dir because I want to use these in `valid-context-access` rule.

src/rules/infinite-reactive-loop.ts
```diff
@@ -400,12 +354,14 @@ export default createRule("infinite-reactive-loop", {
     type: "suggestion",
   },
   create(context) {
+    const tickCallExpressions = Array.from(
+      extractSvelteLifeCycleReferences(context, ["tick"]),
+    )
+    const taskReferences = Array.from(extractTaskReferences(context))
+    const reactiveVariableReferences = getReactiveVariableReferences(context)
```
baseballyama, line 361: Just I moved there lines for performance improvement.

src/rules/valid-context-access.ts
baseballyama: I added this logic and I wrote this limitation on the docs.

src/rules/valid-context-access.ts
```diff
@@ -0,0 +1,223 @@
...
+              }
+            }
+          }
+        } else if (parent?.type === "ExpressionStatement") {
+          if (parent.expression.type !== "CallExpression") {
+            report(contextCallExpression)
+          } else if (lifeCycleReferences.includes(parent.expression)) {
+            report(contextCallExpression)
+          }
+        }
```
ota-meshi, line 197: I think it's safer to check the VariableDeclarator. Also, I think we should check if the right operand is a function expression.
Otherwise, the following cases will result in false positives.

```svelte
<script>
  import { getContext, onMount } from "svelte"

  const foo = getContext("foo")

  onMount(() => {
    foo()
  })
</script>

```

```suggestion
        if (
          (parent?.type === "VariableDeclarator" &&
            parent.init &&
            (parent.init.type === "FunctionExpression" ||
              parent.init.type === "ArrowFunctionExpression") &&
            isInsideOf(parent.init, currentNode)) ||
          parent?.type === "FunctionDeclaration"
        ) {
          const references = parent.id ? getReferences(parent.id) : []

          for (const reference of references) {
            if (reference.identifier?.parent?.type === "CallExpression") {
              if (
                !visitedCallExpressions.includes(reference.identifier.parent)
              ) {
                visitedCallExpressions.push(reference.identifier.parent)
                doLint(
                  visitedCallExpressions,
                  contextCallExpression,
                  reference.identifier?.parent,
                )
              }
            }
          }
        } else if (parent?.type === "ExpressionStatement") {
          if (parent.expression.type !== "CallExpression") {
            report(contextCallExpression)
          } else if (lifeCycleReferences.includes(parent.expression)) {
            report(contextCallExpression)
          }
        }
        parent = parent.parent
```

ota-meshi (changes requested):

ota-meshi:
The repository has been migrated to a monorepo.
I'm marking this as a draft because I think it requires major changes.
If anyone wants to implement this, feel free to open another PR.

Issue #476, opened 2023-05-11
svelte/@typescript-eslint/no-unnecessary-condition possible false positive


olafurw:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.

### What version of ESLint are you using?

8.39.0

### What version of `eslint-plugin-svelte` are you using?

2.28.0

### What did you do?

https://typescript-eslint.io/play/#ts=5.0.4&sourceType=module&code=KYDwDg9gTgLgBAG2PAFlYAzAXHAzjKASwDsBzOAHzgFdiATTE4OuAXhvseOYG4AoPgGMIxfDSgI2cNJjgB+OXADkKGDDBYA9JtABDALZgkAOmH6l-IA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK+SpPRRE0aB2iRwYAL4gtQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA

I'm showing here in eslint playground but the same thing happens in a `.svelte` file, with both `svelte/@typescript-eslint/no-unnecessary-condition` and `@typescript-eslint/no-unnecessary-condition`

```
<script>
export let href: string | undefined = undefined;

const url = href ?? 'http://example.com';
</script>
```

It will tell me that the `href` condition check is unnecessary since it will always be null or undefined.
But it's an `export` variable, so it might get some value from whoever is using the property, I've just defaulted it to `undefined` if someone doesn't send it in.

### What did you expect to happen?

No warning/error

### What actually happened?

Warning: `Unnecessary conditional, left-hand side of `??` operator is always `null` or `undefined`. 3:13 - 3:17`

### Link to **GitHub Repo** with Minimal Reproducible Example

[...]

PR #481, opened 2023-05-15
feat: add `svelte/no-loss-of-prop-reactivity` rule

ota-meshi:
This PR adds `svelte/no-loss-of-prop-reactivity` rule that disallow the use of props that potentially lose reactivity.

Related to https://github.com/sveltejs/eslint-plugin-svelte/issues/476#issuecomment-1546832126


jonatansberg:
Just was I was looking for! Would it be possible to extend this to stores as well?

PR #489, opened 2023-05-22
feat: Added the no-unused-class-name rule using parser services

marekdedic:
Alternative proposal to #402, see https://github.com/sveltejs/svelte-eslint-parser/pull/318#issuecomment-1521005109


docs/rules/no-unused-class-name.md
```diff
@@ -0,0 +1,64 @@
+---
+pageClass: "rule-details"
+sidebarDepth: 0
+title: "svelte/no-unused-class-name"
+description: "disallow the use of a class in the template without a corresponding style"
+since: "v2.16.0"
```
ota-meshi, line 6: Please remove it as it will be added automatically.

```suggestion
```

After that you need to run `pnpm run update`.

src/rules/no-unused-class-name.ts
```diff
@@ -0,0 +1,116 @@
...
+    | SvelteAttribute
+    | SvelteShorthandAttribute
+    | SvelteSpreadAttribute
+    | SvelteDirective
+    | SvelteStyleDirective
+    | SvelteSpecialDirective,
+): string[] {
+  if (attribute.type === "SvelteAttribute" && attribute.key.name === "class") {
+    return attribute.value.flatMap((value) =>
+      value.type === "SvelteLiteral" ? value.value.split(" ") : [],
```
ota-meshi, line 78: I think we should use regex because we can also use tabs etc.

```suggestion
      value.type === "SvelteLiteral" ? value.value.split(/\s+/u) : [],
```

src/rules/no-unused-class-name.ts
```diff
@@ -0,0 +1,116 @@
...
+ * Extract all class names used in a PostCSS selector.
+ */
+function findClassesInSelector(node: SelectorNode): string[] {
+  if (node.type === "class") {
+    return [node.value]
+  }
+  if (node.type === "root" || node.type === "selector") {
+    return node.nodes.flatMap(findClassesInSelector)
+  }
+  return []
```
ota-meshi, line 115: I don't think it can handle selectors like `:not(.foo)`, `:is(.foo, .bar)`. Could you please add test cases?

ota-meshi (changes requested):

src/rules/no-unused-class-name.ts
marekdedic: Good catch, fixed & added a test

marekdedic:
There seems to be some error produced by prettier for `tests/fixtures/rules/no-unused-class-name/valid/unknown-lang01-input.svelte`, I haven't figured out how to get rid of it...

ota-meshi (approved):
LGTM! Thank you for many improvements and implementation of this rule.

PR #491, opened 2023-05-28
fix spelling of SvelteKit

benmccann:


ota-meshi (approved):
Thank you very much 😄

PR #497, opened 2023-06-04
chore: remove `require-jsdoc` internal rule

baseballyama:
In my experience in this repository, I felt that in most cases I have to write comments what is known by the function name too, so I thought it would be okay to remove this rule.


ota-meshi (approved):
:+1:

Issue #493, opened 2023-05-30
Rule Proposal: `svelte/component-tags-order`


Shyam-Chen:
### Description

Refs:
- https://eslint.vuejs.org/rules/component-tags-order.html
- https://eslint.vuejs.org/rules/define-macros-order.html

```svelte
<!-- ✓ GOOD -->
<svelte:options />

<script context="module">/* ... */</script>
<script>/* ... */</script>

<svelte:head>/* ... */</svelte:head>

<svelte:window />
<svelte:document />
<svelte:body />

<div>...</div>
<div>...</div>

<style>/* ... */</style>
```

```svelte
<!-- ✓ GOOD -->
<script>
  import { page } from '$app/stores';
</script>

<svelte:head>
  <title>{$page.data.title}</title>
</svelte:head>

<div>...</div>
```

marekdedic:
Hi, let me argue against adding this rule:

This is already covered by `prettier-plugin-svelte` as described in its [docs](https://github.com/sveltejs/prettier-plugin-svelte#svelte-sort-order). Please don't duplicate this.

PR #498, opened 2023-06-04
feat: add `svelte/component-tags-order` rule

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/493


.prettierignore
```diff
@@ -2,4 +2,5 @@
 .type-coverage
 build
 /lib
-.npmrc
\ No newline at end of file
+.npmrc
+tests/fixtures/rules/component-tags-order
```
baseballyama, line 6: How can I disable prettier for specific file?

ota-meshi: I often use `<!-- prettier-ignore -->`, but it is OK to use a `.prettierignore` file.

https://github.com/sveltejs/eslint-plugin-svelte/blob/e386e3417afa30d69e992f56dd8aff046a37c4cf/tests/fixtures/rules/indent/invalid/4-indent/test01-input.svelte#L1

docs/rules/component-tags-order.md
```diff
@@ -0,0 +1,142 @@
...
+
+## :wrench: Options
+
+```json
+{
+  "svelte/component-tags-order": [
+    "error",
+    {
+      "order": [
+        "SvelteScriptElement([context=module])",
```
ota-meshi, line 111: Hmm. I don't think it's a good idea to use a format specifically for this rule.
For example, using the CSS selector format most users already know the format. (However, users may be a little surprised because the colon must be escaped. e.g. `svelte\:options`)
Is the format currently used by the rule inspired by something?

baseballyama: I copied most parts from https://eslint.vuejs.org/rules/component-tags-order.html
And while implementing the rule, I felt that we need to support `and` condition, but I couldn't find the way to use `and` condition in the ESLint selector.
https://eslint.org/docs/latest/extend/selectors

And I think that the ESLint selector is complicated for most users because they need to understand AST.

That's why I go to use a special format.
(But yeah, I didn't support escaping things...)

ota-meshi: > And while implementing the rule, I felt that we need to support `and` condition, but I couldn't find the way to use `and` condition in the ESLint selector.

Are you saying we can't use `and` in CSS selectors?
I think that the and condition in the CSS selector just connects without using spaces. e.g. `script[context=module][lang=ts]`
And in modern CSS selectors the `or` condition uses `:is()`. e.g. `:is(script, style)`


> https://eslint.org/docs/latest/extend/selectors

I agree that esquery is difficult for users to understand.

ota-meshi: By the way, eslint-plugin-vue implements a function that converts CSS selectors to matchers in order to use CSS selectors in defining [`vue/component-tags-order`](https://eslint.vuejs.org/rules/component-tags-order.html) rule option.



https://github.com/vuejs/eslint-plugin-vue/blob/81ce0cecb87eae60830b9b973790d4f2386062c1/lib/utils/selector.js#L22

marekdedic:
Hi,
to add context, please see https://github.com/sveltejs/eslint-plugin-svelte/issues/493#issuecomment-1913593690

ota-meshi:
The repository has been migrated to a monorepo.
I'm marking this as a draft because I think it requires major changes.
If anyone wants to implement this, feel free to open another PR.

baseballyama:
Let's delegate it to `prettier-plugin-svelte`.

Issue #436, opened 2023-04-08
Disallow certain HTML elements from being used


dkrutsko:
### Motivation

Svelte can be used as a strongly component-based system where most of the elements used are components. Therefore, it makes sense that an error is raised whenever an element is specified but should not be used.

### Description

Disallows certain built-in HTML elements from being used, instead there's likely a component that should be used in its place. For example, instead of using the `h1` element, import and use the `Heading` element.

### Examples

```svelte
<script>
import Heading from 'heading.svelte';
import Code from 'code.svelte';
</script>

<!-- ✓ GOOD -->
<!-- [ 'error', { disallow: [ 'h1', 'code' ] } ] -->
<div />
<Heading type={1}>H1</Heading>
<Code>code</Code>

<!-- ✗ BAD -->
<h1>H1</h1>
<code>code</code>
```

ota-meshi:
Thank you for positing issue.
That rule sounds good to me.
I think the new rule is similar to the `vue/no-restricted-html-elements` rule, so its implementation may be helpful.
https://eslint.vuejs.org/rules/no-restricted-html-elements.html

I think you can also use the no-restricted-syntax rule if you want to get it right now.
https://eslint.org/docs/latest/rules/no-restricted-syntax

```json
  "rules": {
    "no-restricted-syntax": ["error", "SvelteElement[kind='html'] > SvelteName.name[name=/^(?:h1|code)$/]"]
  }
```

dkrutsko:
@ota-meshi This is a good reference, what I would add, however, is the ability to specify an array of elements when using the `{ element, message }` syntax.

```js
{
	'svelte/no-restricted-html-elements':
	[
		'error',
		{
			element: [ 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ],
			message: 'Prefer use of our custom <Heading /> component'
		}
	]
}
```

ota-meshi:
Welcome pull requests.

PR #499, opened 2023-06-04
feat: add `svelte/no-restricted-html-elements` rule

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/436


docs/rules/no-restricted-html-elements.md
```diff
@@ -0,0 +1,102 @@
...
+
+## :wrench: Options
+
+This rule takes a list of strings, where each string is an HTML element name to be restricted:
+
+```json
+{
+  "svelte/no-restricted-html-elements": [
+    "error",
+    ["h1", "h2", "h3", "h4", "h5", "h6"]
```
ota-meshi, line 76: I prefer flattening over using nested arrays here.
What do you think about flattening?

src/rules/no-restricted-html-elements.ts
```diff
@@ -0,0 +1,62 @@
+import { createRule } from "../utils"
+
+export default createRule("no-restricted-html-elements", {
+  meta: {
+    docs: {
+      description: "disallow specific HTML elements",
+      category: "Extension Rules",
```
ota-meshi, line 7: Hmm. I think the better category for this rule is `Stylistic Issues` among the categories we currently have, but we may need to create a new category 🤔

docs/rules/no-restricted-html-elements.md
baseballyama: I forgot to update this. I updated!

src/rules/no-restricted-html-elements.ts
baseballyama: Ah, I misunderstood the meaning of `Extension Rules`.
Just now I changed the category.
Should we refactor categories in the next major version?

ota-meshi: It looks like we need to run `pnpm run update`.

ota-meshi: > Should we refactor categories in the next major version?

We can always make changes if we have ideas for categorization that make it easier for users to understand 😉

docs/rules/no-restricted-html-elements.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+Alternatively, the rule also accepts objects.
+
+```json
+{
+  "svelte/no-restricted-html-elements": [
+    "error",
+    {
+      "elements": "h1", "h2", "h3", "h4", "h5", "h6",
```
ota-meshi, line 93:
```suggestion
      "elements": ["h1", "h2", "h3", "h4", "h5", "h6"],
```

baseballyama: 🙏

src/rules/no-restricted-html-elements.ts
baseballyama: I thought about it a bit, but I think `Stylistic Issues` is fine for now.

ota-meshi (approved):
LGTM! Thank you!

Issue #508, opened 2023-06-18
Error on `pnpm install`: `Was not able to set git hooks.`


marekdedic:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.

### What version of ESLint are you using?

N/A

### What version of `eslint-plugin-svelte` are you using?

main

### What did you do?

```sh
$ pnpm install
```

### What did you expect to happen?

No error

### What actually happened?

```
❯ pnpm install
Already up to date
Progress: resolved 473, reused 430, downloaded 0, added 0, done

> eslint-plugin-svelte@2.30.0 prepare /home/user/eslint-plugin-svelte
> simple-git-hooks

[ERROR], Was not able to set git hooks. Error: [ERROR] Config was not found! Please add `.simple-git-hooks.js` or `simple-git-hooks.js` or `.simple-git-hooks.json` or `simple-git-hooks.json` or `simple-git-hooks` entry in package.json.
Check README for details
 WARN  Issues with peer dependencies found
.
├─┬ @1stg/browserslist-config 1.2.4
│ └── ✕ missing peer browserslist@>=4.0.0
├─┬ markdown-it-anchor 8.6.7
│ └── ✕ missing peer markdown-it@"*"
├─┬ @1stg/stylelint-config 5.1.0
│ └─┬ stylelint-config-prettier 9.0.5
│   └── ✕ unmet peer stylelint@">= 11.x < 15": found 15.8.0
└─┬ svelte-i18n 3.6.0
  └── ✕ unmet peer svelte@^3.25.1: found 4.0.0-next.1
Peer dependencies that should be installed:
  browserslist@>=4.0.0  markdown-it@"*"

Done in 1.8s
```

### Link to **GitHub Repo** with Minimal Reproducible Example

N/A

PR #544, opened 2023-07-17
chore: remove `simple-git-hooks`

baseballyama:
It was not used at all.😇

close: https://github.com/sveltejs/eslint-plugin-svelte/issues/508


ota-meshi (approved):
Thanks!

PR #567, opened 2023-08-22
docs: update README to use Svelte style guide

benmccann:
Projects in the Svelte org use single quotes for JavaScript strings and our new project creation wizards like `npm init svelte` use single quotes for JavaScript strings as well. Using single quotes in the README would make it a lot easier to setup this plugin because I could copy/paste the code from the README into a project without having to update the formatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ module.exports = {
   extends: [
     // add more generic rule sets here, such as:
     // 'eslint:recommended',
-    "plugin:svelte/recommended",
+    'plugin:svelte/recommended',
   ],
   rules: {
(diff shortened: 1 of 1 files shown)


ota-meshi:
Thank you for this PR! However, I think we need to change the `prettier` config and change the overall source code format of the repository. Is it possible for you to do that?
If it is difficult, I will do it when I have time 😉

benmccann:
yes, good idea! I didn't realize the readme was being formatted. done here: https://github.com/sveltejs/eslint-plugin-svelte/pull/568

PR #608, opened 2023-11-03
feat: added the `no-inline-styles` rule

### LINKED ISSUE #603: Add rule to disallow inline styles
### Motivation

When using CSP, you often want to disallow inline styles (to not have `style-src: 'unsafe-inline'`), I'd like to add a rule that checks this automatically

### Description

The rule would look for any directive producing inline styles and error on them.

Additionaly, using svelte transitions generates inline styles (although that may change in the future if/when svelte migrates to teh web animations - see sveltejs/svelte#6662), so this rule should probably ban using them altogether (gated with a configuration option probably?)

### Examples

```svelte
<!-- ✗ BAD -->
<div style="anything" />
<div style:anything={true} />
<div transition:anything />
```

### Additional comments

_No response_

marekdedic:
Closes #603


docs/rules/no-inline-styles.md
```diff
@@ -0,0 +1,69 @@
...
+  "svelte/no-inline-styles": [
+    "error",
+    {
+      "allowTransitions": false
+    }
+  ]
+}
+```
+
+-
```
ota-meshi, line 60: Could you add an explanation about allowTransitions?

ota-meshi (changes requested):

ota-meshi (approved):
LGTM! Thank you!

Issue #587, opened 2023-09-20
Svelte 5 runes support


benmccann:
### Description

There's a few places it'd be helpful such as in linting the Svelte 5 repo and component party examples (https://github.com/matschik/component-party.dev/issues/186)

I don't think there's any rush on this and the syntax may change, but I thought it'd be helpful to have a place to track it

ota-meshi:
MEMO: 

When using `$derived`, there seems to be a problem with TS inference, similar to the problem we had in the past.

https://github.com/sveltejs/svelte-eslint-parser/issues/226

input:

```svelte
<script lang="ts">
type EventInfo = {
  start_at: Date | number
}
let info: EventInfo | null = null

const lightFormat = (i: Date | number) => i.toString()

fetch('/fakeurl').then(() => { info = { start_at: 0 } }).catch(console.error)

const startDate = $derived(info?.start_at ? lightFormat(info.start_at) : null)
const endDate = () => (info?.start_at ? lightFormat(info.start_at) : null)
</script>
 
{startDate}{endDate}
```

error: 

> Unsafe assignment of an `any` value. (@typescript-eslint/no-unsafe-assignment)

Reproduce:

[...]

PR #620, opened 2023-11-20
feat: (experimental) partial support for Svelte v5

ota-meshi:
related to #587

Also, merging this PR will start a pre-release.


ota-meshi:
It seems that the main branch's CI is broken in the first place. I need to fix that first.

ota-meshi:
For some tests that cannot avoid failure using Svelte v5, I use `*-requirements.json` to exclude them from the tests. The causes of these will need to be investigated later.

baseballyama (approved):

PR #624, opened 2023-11-21
feat: add support for `{@snippet}` and `{@render}` in mustache-spacing rule

ota-meshi:
We need to merge https://github.com/sveltejs/eslint-plugin-svelte/pull/620 first.

related to https://github.com/sveltejs/eslint-plugin-svelte/issues/587


tests/fixtures/rules/mustache-spacing/invalid/always-after-expression/snippet-render01-output.svelte
```diff
@@ -0,0 +1,4 @@
+<!-- prettier-ignore -->
+{#snippet foo() }
```
baseballyama, line 2: I don't think we don't need to add space before closing bracket.

```suggestion
{#snippet foo()}
```

The docs also doesn't have a space and I think it's more natural for me.

image

https://svelte-5-preview.vercel.app/docs/snippets

tests/fixtures/rules/mustache-spacing/invalid/always/snippet-render01-output.svelte
```diff
@@ -0,0 +1,4 @@
+<!-- prettier-ignore -->
+{ #snippet foo() }
```
baseballyama, line 2: Also I don't think we need space before `{`.

```suggestion
{#snippet foo()}
```

baseballyama (approved):
Once I made a comment but after that I understood properly so I removed.😇

PR #656, opened 2024-01-03
docs: clarify docs for `svelte/no-target-blank`

benmccann:
The behavior of `allowReferrer` was very confusing to me because I thought it was referring to the `rel` attribute and I did not understand it's actually referring to the `Referer` header being allowed

I was also very confused about the security vulnerability being referred to. I assumed it was related to protecting the user's privacy by not passing the referrer header. When I happened to click the reference link I saw that I had been misunderstanding the purpose of this rule and what exactly it was doing.


ota-meshi (approved):
Thank you!

PR #665, opened 2024-01-11
feat: use synckit to support postcss-load-config v5

### LINKED ISSUE #636: Move to `postcss-load-config` 5.0
This project is using `postcss-load-config` 3.0, but the latest version is 5.0.

In new version we removed Node.js 12—16, improved ESM and TS support, and moved to async API.

JounQin:
close #636


JounQin:
```log
  76:5  error  "postcss-load-config@^5.0.2" is not compatible with "node@^14.17.0||>=16.0.0". Allowed is: "node@>=18.0.0"  node-dependencies/compat-engines
  81:5  error  "synckit@^0.9.0" is not compatible with "node@^14.17.0||>=16.0.0". Allowed is: "node@^14.18.0||>=16.0.0"    node-dependencies/compat-engines
```

Do we want to release a major version for this?

JounQin:
https://github.com/sveltejs/svelte/blob/ab13356008a609158da7493da4758d8d11645ac4/packages/svelte/package.json#L9

I believe we should match `svelte` itself.

ota-meshi:
Thank you for the PR.
I would like to release eslint-plugin-svelte v2.36 after the stable version of svelte v5 is released, and then work on the next major version.

baseballyama:
Even if Svelte 5 is released, I don't think we can drop Node < 18 as long as we continue to support Svelte 4 and Svelte 3.
I think it is a good idea to support only Svelte 5 in the next major version.

JounQin:
I meant this PR should target on the next major version.

The `main` itself is targeting Svelte v5, how can I create a PR for old releases?

cc @ota-meshi

ota-meshi:
There is currently no way to manage multiple versions in this repository 😓

JounQin:
It needs multiple branches. For example https://github.com/mdx-js/eslint-mdx/tree/release-v2.

PR #679, opened 2024-01-28
feat: added the `no-goto-without-base` rule

### LINKED ISSUE #675: Add rule `no-goto-without-base`
### Motivation

In SvelteKit, the `goto` function is used for internal navigation. That means, that you basically almost always want to prefix the url with the `base` path.

### Description

Add a rule that would trigger on any `goto` without a `base` path. However, there are quite some edge cases

### Examples

```svelte
<script>
  import { base } from "$app/paths";
  import { base as whatever } from "$app/paths";
  import { goto } from "$app/navigation";

  <!-- ✓ GOOD -->
  goto(base + "/foo");
  goto(`${base}/foo`)
  goto(whatever + "/foo");
  goto(`${whatever}/foo`)
  goto("https://absolute.url");

  <!-- ✗ BAD -->
  goto("/foo");
  goto("foo/" + base);
  goto(`foo/${base}`);
</svelte>
```

### Additional comments

Is this plugin the right place to put SvelteKit rules?

marekdedic:
Closes #675


marekdedic:
Hi,
this is basically done from my side. However:
- I would really like a review from somebody who knows SvelteKit better than me
- There are probably some exceptions, when you don't want this rule to trigger - if you know of any, I'd add them in, but I could think only of absolute URLs (but since `goto` is not for external URLs, maybe absolute URLs should be prohibited altogether?). They can be added later of course or put behind some config flag.

ota-meshi:
Thank you for this PR! But, I'm not familiar with creating web apps with SvelteKit.
@baseballyama Can you check if what the rules report useful or not?

baseballyama:
Added `goto` interface improvements in kit 3.0 milestones.
https://github.com/sveltejs/kit/issues/11803

Until then, I think this rule is useful but I think it not `plugin:svelte/recommended` rule because part of users doesn't use basepath.

marekdedic:
Hi,
thanks, I thought about why SvelteKit just doesn't do this by default, but it feels like a quite big breaking change, so I didn't propose it to kit directly, maybe I should've...

I don't think this needs to be recommended, but actually if you don't use basepath, prepending it is still correct in the sense that it doesn't break anything and enables you to use a basepath down the line. But if this is something that would be added to SvelteKit itself down the line, it doesn't really make sense to me to make this recommended either...

baseballyama:
In my opinion, anyway, ESLint allows users to set rules freely, so as long as the rules are not annoying, they can be merged.
I will check this PR will in this weekend. (Sorry for the delay but I’m bit busy for my work now🙏)

src/rules/no-goto-without-base.ts
```diff
@@ -0,0 +1,125 @@
...
+				goto: {
+					[ReferenceTracker.CALL]: true
+				}
+			}
+		}),
+		({ node }) => node
+	);
+}
+
+function extractBasePathNames(referenceTracker: ReferenceTracker): string[] {
```
ota-meshi, line 113: Could you make sure to track the variables and references as below so that we can check the complete references?
Maybe we should change the function name as well.

```ts
import { findVariable } from '../utils/ast-utils';

function extractBasePathNames(
	referenceTracker: ReferenceTracker,
	context: RuleContext
): Set<TSESTree.Identifier> {
	const set = new Set<TSESTree.Identifier>();
	for (const { node } of referenceTracker.iterateEsmReferences({
		'$app/paths': {
			[ReferenceTracker.ESM]: true,
			base: {
				[ReferenceTracker.READ]: true
			}
		}
	})) {
		const variable = findVariable(context, (node as TSESTree.ImportSpecifier).local);
		if (!variable) continue;
		for (const reference of variable.references) {
			if (reference.identifier.type === 'Identifier') set.add(reference.identifier);
		}
	}
	return set;
}
```

ota-meshi (changes requested):

ota-meshi:
Maybe this rule should also check `<a href="...">? What do you think?

src/rules/no-goto-without-base.ts
marekdedic: Thanks, added :)

marekdedic:
> Maybe this rule should also check `<a href="...">`? What do you think?

Well, that's a little bit more complicated. That is because as per the SvelteKit docs, you should use `goto()` for internal navigation (where it always makes sense to check for the base path). `<a>` links can be use for both internal as well as external navigation. We could, however, maybe check all `<a>` elements and report missing base iff the path is relative...

ota-meshi (approved):
Thank you for updating this PR! LGTM!
Also, thank you for your opinion. Let's make checking for the `<a>` tag a separate rule, as different users may have different preferences.
(I fixed the CI error in another PR.)

Issue #689, opened 2024-02-26
Add option to svelte/prefer-class-directive - apply only when one side is the empty string


sdarnell:
### Description

The examples for prefer-class-directive show a clear benefit for using the directive form. That is:
```
BAD : class={current === 'foo' ? 'selected' : ''}
GOOD: class:selected={current === 'foo'}
```

But there are other scenarios where this rule is far less obviously a good thing, for example:
```html
BAD ?: class="btn {isDanger ? 'btn-danger' : 'btn-primary'}" 
GOOD?: class="btn" class:btn-danger={isDanger} class:btn-primary={!isDanger} 
```

Personally I prefer the class directive form (which is why i have this rule enabled), and the version preferred by the rule avoids direct interpolation into class, but is longer (more verbose), and uses the condition twice, which in this example is a simple boolean, but might be a bit more complex.
The option using the ternary also emphasises that we only use one *or* the other class name, whereas I'd argue it is a bit less clear that they are exclusive when using directives.

So my feature request is to add a configuration option to only enforce class directives when one of the options is the empty string (the rule example).  The goal being that the rule can stay enabled, but be less dogmatic for the pretty common case of 'this or that class'.  You may find that it could be moved to the recommended set with the proposed option.

Or is there some other justification which makes the ternary approach a bad idea?
If so, I'd suggest adding this reasoning to the rule description.

ota-meshi:
Thank you for suggesting options.  Sounds good to me.  Pull requests are welcome.

sdarnell:
@ota-meshi OK, great, thanks. I've prepared PR #690.

PR #690, opened 2024-02-27
feat: Add prefer option to prefer-class-directive

sdarnell:
This the prefer option can be 'empty' or 'always' with 'empty' being the default. 'always' matches the current behaviour, always insisting on directives 'empty' only prefers directives when one class is empty.

### DIFF
--- a/docs/rules/prefer-class-directive.md
+++ b/docs/rules/prefer-class-directive.md
@@ -22,13 +22,16 @@ This rule aims to replace a class with ternary operator with the class directive
 
 ```svelte
 <script>
-  /* eslint svelte/prefer-class-directive: "error" */
+  /* eslint svelte/prefer-class-directive: "empty" */
+  const selected = true;
 </script>
 
 <!-- ✓ GOOD -->
+<button class:selected={selected}>foo</button>
 <button class:selected={current === 'foo'}>foo</button>
 
 <!-- ✗ BAD -->
+<button class={selected ? 'selected' : ''}>foo</button>
 <button class={current === 'foo' ? 'selected' : ''}>foo</button>
 ```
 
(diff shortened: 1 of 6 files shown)


sdarnell:
Fixes #689

docs/rules/prefer-class-directive.md
```diff
@@ -22,13 +22,16 @@ This rule aims to replace a class with ternary operator with the class directive
 
 ```svelte
 <script>
-  /* eslint svelte/prefer-class-directive: "error" */
+  /* eslint svelte/prefer-class-directive: "empty" */
```
ota-meshi, line 25: I think the configuration is probably wrong.

```suggestion
  /* eslint svelte/prefer-class-directive: ["error", {"prefer": "empty"}] */
```

tests/fixtures/rules/prefer-class-directive/invalid/_config.json
```diff
@@ -0,0 +1,3 @@
+{
+	"options": [{ "prefer": "always" }]
+}
```
ota-meshi, line 3: Could you please add test cases to check what is reported as an error when using `empty`?

ota-meshi (changes requested):
Thank you for this PR!

docs/rules/prefer-class-directive.md
sdarnell: Fixed. Thanks.

tests/fixtures/rules/prefer-class-directive/invalid/_config.json
sdarnell: I've added some specific tests, expanding on test 01.
But note that as `prefer: "empty"` is a relaxation of the rule, the other tests cover it quite well.
While doing that I noticed that the 'empty' rule should also apply when the classname contains just whitespace - so I rejigged my changes to handle that too.

ota-meshi (approved):
Thank you! LGTM!

PR #735, opened 2024-04-10
feat: add rule types

### LINKED ISSUE #707: Ship types for `eslint-define-config`
### Description

https://github.com/eslint-types/eslint-define-config

Ship types to provide autocompletion and documentation for rules in their `defineConfig` function.

I would be willing to create a PR for this and maintain the types if no one else wants to.

xiBread:
Fixes #707 

This PR deviates slightly from the original request; instead of supporting `eslint-define-config`, this directly augments ESLint's `Linter.FlatConfig` type:

Screenshot 2024-04-10 at 7 14 42 PM

This relies on [eslint-typegen](https://github.com/antfu/eslint-typegen). ~~I've added a new build script in order to preserve the comments in this `rule-types.d.ts` file since `tsconfig.build.json` removes them (if there's a better way to do this then please let me know).~~


package.json
```diff
@@ -27,9 +27,10 @@
     "sveltejs"
   ],
   "scripts": {
-    "build": "pnpm run build:meta && pnpm run build:ts",
+    "build": "pnpm run build:meta && pnpm run build:ts && pnpm run build:types",
     "build:meta": "pnpm run ts ./tools/update-meta.ts",
     "build:ts": "tsc --project ./tsconfig.build.json",
+	"build:types": "pnpm run ts ./tools/update-rule-types.ts && copyfiles -f ./src/rule-types.d.ts ./lib",
```
ota-meshi, line 33: Do we need a copy of the file? Doesn't the copied file get created just by running `tsc`?

ota-meshi (changes requested):
Thank you for the PR! I didn't know we could do that with type definitions.

src/rule-types.d.ts
```diff
@@ -0,0 +1,579 @@
+// IMPORTANT!
+// This file has been automatically generated,
+// in order to update its content execute "pnpm run update"
+
+/* eslint-disable */
```
ota-meshi, line 5: It looks like we need to add this file to `.eslintigore`.

package.json
ota-meshi: Ah, since `tsdoc` is removed, we are copying the file.
Let's remove `removeComments` from `tsconfig.build.json`.

src/rule-types.d.ts
ota-meshi, line 1: Sorry, it seems that unless we make the file a simple `.ts` file, `tsc` will not generate it as a compiled file.
Could you please rename the file to `rule-types.ts`?

ota-meshi (changes requested):

package.json
```diff
@@ -27,9 +27,10 @@
     "sveltejs"
   ],
   "scripts": {
-    "build": "pnpm run build:meta && pnpm run build:ts",
+    "build": "pnpm run build:meta && pnpm run build:ts && pnpm run build:types",
     "build:meta": "pnpm run ts ./tools/update-meta.ts",
     "build:ts": "tsc --project ./tsconfig.build.json",
+    "build:types": "pnpm run ts ./tools/update-rule-types.ts",
```
ota-meshi, line 33: I don't think this is necessary since it is generated in the `update` script.

ota-meshi (approved):
LGTM! Thank you!

PR #741, opened 2024-04-23
feat: Follow latest Svelte 5

baseballyama:
Use the latest Svelte 5 compatible parser.

Should we increase the minor version for this change?


ota-meshi (approved):
Thank you!

> Should we increase the minor version for this change?

Yeah, I think it's safer to increase a minor version.

PR #749, opened 2024-05-06
feat: add `svelte/no-svelte-internal` rule

baseballyama:
`svelte/internal` is deprecated in Svelte 5 and will be removed in Svelte 6 according to our current plan.
The Svelte compiler only compiles `.svelte` files and `.svelte.[jt]s` files, so the compiler can not check normal `.[jt]s` files. Therefore, the ESLint plugin checks this.


baseballyama:
I forgot to write a document.😇

baseballyama:
Finished ~~ @ota-meshi

src/rules/no-svelte-internal.ts
```diff
@@ -0,0 +1,33 @@
...
+		},
+		schema: [],
+		messages: {
+			unexpected: 'Importing from svelte/internal is prohibited. This will be removed in Svelte 6.'
+		},
+		type: 'problem'
+	},
+	create(context) {
+		return {
+			ImportDeclaration(node) {
```
ota-meshi, line 20: We should probably also check `export * from 'svelte/internal'`.

ota-meshi (changes requested):

src/rules/no-svelte-internal.ts
ota-meshi: I don't expect anyone to do that, but I think it would be even better if we also check `inport('svelte/internal')`.

baseballyama: Good! updated: https://github.com/sveltejs/eslint-plugin-svelte/pull/749/commits/fca148ac1273256aea7cf656f9015d481b3580f8

ota-meshi (approved):
LGTM! Thank you!

PR #750, opened 2024-05-06
chore: add Node22 in CI

baseballyama:


baseballyama:
I think I need to make a PR to fix these errors.

ota-meshi (approved):
Thank you!

ota-meshi:
Hmm. CI is failing on the main branch.
eslint-plugin-n was released a few minutes ago so that might be the cause. I'll look into it later.

https://www.npmjs.com/package/eslint-plugin-n?activeTab=versions

PR #752, opened 2024-05-07
fix: disable a few rules in docs-svelte-kit

baseballyama:
I have no idea how to fix below lint errors.
https://github.com/sveltejs/eslint-plugin-svelte/actions/runs/8978756715/job/24659835714

So I just disable rules.


ota-meshi (approved):
Thanks so much!

PR #770, opened 2024-06-08
chore: Isolate the cause of ghpages publishing failure.

ota-meshi:


ota-meshi:
There seems to be a problem with the GHPages pipeline, and I've tried to run this pull request to identify it, but it seems difficult.
First, I'll break it down into steps to make it easier to identify the problem.

baseballyama:
You may know this, but just in case. I already pushed some commit.
https://github.com/sveltejs/eslint-plugin-svelte/commit/d11f282e796716a79433969581bc278c719ae99d

ota-meshi:
Wow, it was already fixed, thanks.

PR #772, opened 2024-06-09
fix: false positive for kebab-case with svelte v5 in `svelte/no-unused-svelte-ignore`

### LINKED ISSUE #765: Old-styled svelte-ignores are ignored with Svelte 5

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have updated to the latest version of the packages.

9.3.0

2.39.0

### What did you do?

<details>
<summary>Configuration</summary>

I use default configuration provided by `npm create svelte@latest`
```
import js from '@eslint/js';
import svelte from 'eslint-plugin-svelte';
import globals from 'globals';

/** @type {import('eslint').Linter.FlatConfig[]} */
export default [
	js.configs.recommended,
	...svelte.configs['flat/recommended'],
	{
		languageOptions: {
			globals: {
				...globals.browser,
				...globals.node
			}
		}
	},
	{
		ignores: ['build/', '.svelte-kit/', 'dist/']
	}
];

```
</details>

```svelte
<!-- svelte-ignore a11y-click-events-have-key-events a11y-no-static-element-interactions -->
<my-tag on:click={() => {}}>
</my-tag>

<div />

```


Like when Svelte 4.2.17 is used, ` a11y-click-events-have-key-events` and `a11y-no-static-element-interactions` should not be reported.

Also `element_invalid_self_closing_tag` was also not reported with Svelte 4, so IMO should not be reported in legacy mode with Svelte 5.

### What actually happened?

```
eslint-plugin-svelte-repro/src/routes/+page.svelte
  1:20  error  svelte-ignore comment is used, but not warned                                                                                                                                                                                                                                                                                                                svelte/no-unused-svelte-ignore
  1:54  error  svelte-ignore comment is used, but not warned                                                                                                                                                                                                                                                                                                                svelte/no-unused-svelte-ignore
  2:1   error  Visible, non-interactive elements with a click event must be accompanied by a keyboard event handler. Consider whether an interactive element such as `<button type="button">` or `<a>` might be more appropriate. See https://svelte.dev/docs/accessibility-warnings#a11y-click-events-have-key-events for more details(a11y_click_events_have_key_events)  svelte/valid-compile
  2:1   error  `<my-tag>` with a click handler must have an ARIA role(a11y_no_static_element_interactions)                                                                                                                                                                                                                                                                  svelte/valid-compile
  5:1   error  Self-closing HTML tags for non-void elements are ambiguous — use `<div ...></div>` rather than `<div ... />`(element_invalid_self_closing_tag)                                                                                                                                                                                                               svelte/valid-compile

✖ 5 problems (5 errors, 0 warnings)
```

https://github.com/shadow-identity/eslint-plugin-svelte-repro

### Additional comments

I think the reason of the mess is that svelte 5 uses underscores in warning names. But according to Svelte developers, old-styled components (which does not use runes) should keep working: https://github.com/sveltejs/svelte/issues/11666

After https://github.com/sveltejs/svelte/issues/11666 was fixed, `svelte-check` does not report this markup.

I also put `<div />` case, it was also not reported with Svelte 4 (it is reported with `svelte-kit` tho).

ota-meshi:
fixes #765


baseballyama (approved):
Thank you!

PR #784, opened 2024-06-14
chore: update deps and fix testcase

ota-meshi:


packages/eslint-plugin-svelte/tests/fixtures/rules/indent/invalid/ts/ts-function02-input.svelte
```diff
@@ -1,7 +1,6 @@
 <!-- prettier-ignore -->
 <script lang="ts">
 declare function
-*
```
ota-meshi, line 4: It seems that an update to the ts parser caused the previous statement to become a syntax error.

PR #789, opened 2024-06-15
chore: Use eslint types for exported configs

KuSh:
This allows svelte configs to pass typescript-eslint type checking

Actually when migrating a project to eslint 9 I get typescript 2345 error
```ts
Argument of type '{ plugins: { readonly svelte: Plugin; }; files?: undefined; languageOptions?: undefined; rules?: undefined; processor?: undefined; } | { files: string[]; languageOptions: { parser: any; }; rules: { ...; }; processor: string; plugins?: undefined; } | { ...; }' is not assignable to parameter of type 'ConfigWithExtends'.
  Type '{ files: string[]; languageOptions: { parser: any; }; rules: { 'no-inner-declarations': string; 'no-self-assign': string; 'svelte/comment-directive': string; 'svelte/system': string; }; processor: string; plugins?: undefined; }' is not assignable to type 'ConfigWithExtends'.
    Types of property 'rules' are incompatible.
      Type '{ 'no-inner-declarations': string; 'no-self-assign': string; 'svelte/comment-directive': string; 'svelte/system': string; }' is not assignable to type 'Partial<Record<string, RuleEntry>>'.
        Property ''no-inner-declarations'' is incompatible with index signature.
          Type 'string' is not assignable to type 'RuleEntry | undefined'. ts(2345)
---
(property) 'flat/recommended': ({
    plugins: {
        readonly svelte: ESLint.Plugin;
    };
    files?: undefined;
    languageOptions?: undefined;
    rules?: undefined;
    processor?: undefined;
} | {
    files: string[];
    languageOptions: {
        parser: any;
    };
    rules: {
        ...;
    };
    processor: string;
    plugins?: undefined;
} | {
    ...;
})[]
```

Using const to define rule levels fixe the problem


packages/eslint-plugin-svelte/src/configs/all.ts
```diff
@@ -6,7 +6,7 @@ export = {
 	extends: [baseExtend],
```
baseballyama, line 6: Could this be addressed by specifying the type? It seems challenging to use `as const` extensively, so I thought this might be a simpler solution.

```diff
- export = {
+ import type { Linter } from 'eslint';
+ const config: Linter.Config = { ... };
+ export = config;
```

KuSh: Sure I can try that top !

KuSh: Didn't see most configs were generated, I've also fixed that

baseballyama:
@KuSh Thank you for the update! Can you add the changeset?

KuSh:
> @KuSh Thank you for the update! Can you add the changeset? I think this can be recognize as fix.

Done !

baseballyama (approved):

ota-meshi (approved):
Thank you!

jtbandes:
Hi there, on 2.43.0, and with `@types/eslint` v9, I'm still seeing an error when trying to type-check my eslint.config.mjs:

image

It looks like this is because the generated file `eslint-plugin-svelte/lib/index.d.ts` is as follows:

```ts
declare const _default: {
    meta: typeof processor.meta;
    configs: {
        base: import("eslint").Linter.Config<import("eslint").Linter.RulesRecord, import("eslint").Linter.RulesRecord>;
        recommended: import("eslint").Linter.Config<import("eslint").Linter.RulesRecord, import("eslint").Linter.RulesRecord>;
        prettier: import("eslint").Linter.Config<import("eslint").Linter.RulesRecord, import("eslint").Linter.RulesRecord>;
        all: import("eslint").Linter.Config<import("eslint").Linter.RulesRecord, import("eslint").Linter.RulesRecord>;
        'flat/base': import("eslint").Linter.FlatConfig<import("eslint").Linter.RulesRecord>[];
        'flat/recommended': import("eslint").Linter.FlatConfig<import("eslint").Linter.RulesRecord>[];
        'flat/prettier': import("eslint").Linter.FlatConfig<import("eslint").Linter.RulesRecord>[];
        'flat/all': import("eslint").Linter.FlatConfig<import("eslint").Linter.RulesRecord>[];
    };
...
```


You can see that TypeScript considers these types invalid:
image

- indeed it only takes 1 generic arg: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7599080ecdafbbe188c857c19a177bc7fe562a4e/types/eslint/index.d.ts#L1222

image

- defined as a non-generic type alias: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7599080ecdafbbe188c857c19a177bc7fe562a4e/types/eslint/index.d.ts#L1281

KuSh:
@jtbandes I don't have any problem with latest versions of both eslint and eslint-plugin-svelte (even if type definitions could be simplified)

PR #796, opened 2024-06-17
feat: improve `svelte/valid-compile` to use `svelte.config.js`'s `onwarn`

### LINKED ISSUE #311: Ignore specific warnings of svelte-check
### Description

For `eslint-plugin-svelte3` there is `svelte3/compiler-options` or `svelte3/ignore-warnings` to ignore some specified compiler warnings, for example all the a11y warnings.

For this plugin I can't seem to find something similar. I can disable the `valid-compile` rule all together, yet that's a bit unspecific. I could try to disable the warning in my svelte.config.js by defining a  `config.onwarn` function but this does not seem to work with vite/kit.

I suggest to provide a way to ignore specific warnings of the svelte compiler, as I think it might be quite handy.

ota-meshi:
close #311


baseballyama (approved):
Nice! Lint is failed.

PR #797, opened 2024-06-17
fix: `svelte/valid-compile` to use verbatimModuleSyntax to work with TS v5.5.

ota-meshi:
See comment https://github.com/sveltejs/eslint-plugin-svelte/pull/790#discussion_r1642834898


baseballyama (approved):
I learned this option😅 and confirmed the behavior on TypeScript playground.

PR #805, opened 2024-06-23
fix: update `svelte-eslint-parser` to fix nested `{#snippet}`

baseballyama:


ota-meshi:
Hmm. The latest `svelte/compiler` v5 seems to give an OOM error for the following code. I don't know why yet.

```svelte
{#snippet
foo(
{
a
}
)
}
{/snippet}
{@render
foo(
{
a
}
)
}
```

ota-meshi:
There seems to be no problem with 5.0.0-next.163. There may be a problem with the difference with 5.0.0-next.164.

ota-meshi:
I haven't written a test case yet, so I'm not sure, but we may need to add the s flag to handle newline characters.

https://github.com/sveltejs/svelte/pull/12070/files#diff-07a9d6689d63c982b2e3a45ae4bd624240d27824e8c5bdce8a539b4d2a7c36f1R282

ota-meshi:
It seems to have been fixed already, we just need to wait for the next version release.

https://github.com/sveltejs/svelte/pull/12144/files

.github/workflows/NodeCI.yml
```diff
@@ -76,7 +76,9 @@ jobs:
         run: pnpm install -D -w eslint@${{ matrix.eslint }}
 
       - name: Test
-        run: pnpm test
+        run: |
+          export NODE_OPTIONS="--max-old-space-size=8192"
```
ota-meshi, line 80: I expect this workaround will no longer be necessary once v5.0.0-165 is released.

ota-meshi:
Some issues were still not fixed, so I opened an issue in the svelte repo.

https://github.com/sveltejs/svelte/issues/12156

ota-meshi (approved):
Thank you!
Everything works fine with the latest version of svelte/compiler.

PR #806, opened 2024-06-26
feat: signal-prefer-let rule

bfanger:
Svelte 5 allows using both `let` and `const` for signals, in regular JavaScript `const` means:

> The variable identifier cannot be reassigned.
> You should understand const declarations as "create a variable whose identity remains constant"
> 

But signals do change and are reassigned by Svelte's reactivity system.

This ESLint rule converts:
```ts
const { value } = $props()
```
into
```ts
let { value } = $props()
```


baseballyama:
Thank you for the PR!

This is just small point but return value of `$effect.root` should use `const` according to the docs, so I think `signal-declaration` or something like that is better rule name and we can check `$effect.root` also.

```js
	const cleanup = $effect.root(() => {
		$effect(() => {
			console.log(count);
		});

		return () => {
			console.log('effect root cleanup');
		};
	});
```

https://svelte-5-preview.vercel.app/docs/runes#$effect-root

ota-meshi:
Thank you for the rule suggestion and opening this PR!
However, I think this rule will conflict with the `prefer-const` rule in ESLint core as it is.
I think we need to consider how to avoid the conflict.
Could you please propose the rule in a new issue first?

bfanger:
@ota-meshi I was thinking about implementing a `svelte/prefer-const` rule which  would extend `prefer-const` but ignore signals declarations.

@baseballyama `$effect.root` is a _rune_, but not a _signal_, the `svelte/prefer-const` rule should convert it to a const

baseballyama:
`$derived` is also rune. Not signal. So I think we need to use `rune` instead of `signal` for the rule name.

bfanger:
`$derived` returns a signal: https://github.com/sveltejs/svelte/blob/8904e0f6cca7c423cd6232ba38870362cf840e63/packages/svelte/src/internal/client/reactivity/deriveds.js#L23

If the svelte compiler injects `$.get()` when you're using the variable it's a signal.

baseballyama:
I know. But we don't use the word "signals" for Svelte developers. We use "runes" instead. We need to use word which mentioned in the docs as much as possible to reduce cognitive cost.

justingolden21:
Just wanted to say thanks for working on this! I've got hundreds of "errors" from eslint about using const instead but I destructure props and some props are consts and some aren't. I'd rather save the red lines for actual real problems. This change would be huge if/when merged. Thanks @bfanger !

bfanger:
@justingolden21  When using eslint-plugin-svelte  3.0.0-next.9 or newer you can use the new "svelte/prefer-const" rule:

```js
rules: {
  "prefer-const": "off", // <- only needed when another preset enables the prefer-const rule.
  "svelte/prefer-const": ["warn", { destructuring: "all" }],
}
```

PR #807, opened 2024-06-27
Add rule for prevent from printing undefined to HTML

### LINKED ISSUE #747: New rule: `no undefined variable in the DOM`
### Motivation

When using a variable (say a props) that can be `null` or `undefined` directly in the DOM, it will be rendered as a string `"null"` or `"undefined"`. Hence, it is probablly never going to be a good thing to use a possibly null or undefined variable directly in the DOM.

### Description

The rule would highlight any variable or expression used in the DOM that resolves to a type involving a union type to `null` or `undefined`.

### Examples

```svelte
<script lang="ts">
  export let prop: string | undefined = undefined;

  const frameworks = ["svelte", "react"]
</script>

<!-- ✓ GOOD -->
{#if prop}
  {prop}
{/if}

<!-- ✗ BAD -->
{prop}
{frameworks.find(framework => framework === prop)}
```

### Additional comments

_No response_

DongwonTTuna:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/747


packages/eslint-plugin-svelte/src/rules/no-undefined-print.ts
```diff
@@ -0,0 +1,68 @@
...
+export default createRule('no-undefined-print', {
+	meta: {
+		docs: {
+			description: 'Disallow from printing `undefined`',
+			category: 'Possible Errors',
+			recommended: true
+		},
+		schema: [],
+		messages: {
+			unexpected: 'Unexpected `undefined`.'
```
baseballyama, line 13:
```suggestion
			unexpected: 'Disallow printing `undefined`'
```

packages/eslint-plugin-svelte/tests/fixtures/rules/no-undefined-print/invalid/html-input.svelte
```diff
@@ -0,0 +1,5 @@
+<script>
+	let string = `this string contains some <strong>HTML!!!</strong>`;
+</script>
+
+<p>{@html string}</p>
```
baseballyama, line 5: I think this test is not for this rule.

baseballyama (changes requested):
Thank you for your PR!

But it seems to me that the rule cannot be made great without a check using types such as this comment. And I think we need to handle null also.

https://github.com/sveltejs/eslint-plugin-svelte/issues/747#issuecomment-2093862131

And there are several points that clearly need to be corrected.

PR #816, opened 2024-07-06
feat: Add svelte/prefer-const rule that excludes reactive variables

bfanger:
> The variable identifier cannot be reassigned.
> You should understand const declarations as "create a variable whose identity remains constant"
>

The ESLint prefer-const doesn't take into account that Svelte's reactivity system is reassigning reactive variables.
The `svelte/prefer-const` rule skip variables that are created by runes like `$props`, `$derived`, `$derived.by`, etc.

Companion to the [rune-prefer-let](https://github.com/sveltejs/eslint-plugin-svelte/pull/806) rule.


ota-meshi:
Could you please propose the rule in a new issue first?

mikededo:
@ota-meshi and @baseballyama, we could close this PR already.

baseballyama:
Close in favor of https://github.com/sveltejs/eslint-plugin-svelte/pull/933

PR #841, opened 2024-08-10
feat: add config option for foreign elements in `svelte/html-self-closing` rule

### LINKED ISSUE #837: More configuration options for `html-self-closing`
### Description

I would like to be able to configure `html-self-closing` so that it aways produces most compact, yet correct tags. That would mean preferring `void` > `self-closing` > `separate closing tag`, but only within spec limitations.

I was able to accomplish that for HTML/Svelte with `{ void: 'never', normal: 'never' }` settings (example output: `<div></div><Component/>`), but now it changes SVG tags from `<path/>` to `<path></path>`.

jrmajor:
This PR adds separate configuration option for foreign (SVG and MathML) elements to `svelte/html-self-closing` rule.

According to HTML spec (https://html.spec.whatwg.org/multipage/syntax.html#elements-2):
> Raw text, escapable raw text, and normal elements have a start tag to indicate where they begin, and an end tag to indicate where they end. The start and end tags of certain normal elements can be omitted, as described below in the section on optional tags. Those that cannot be omitted must not be omitted. Void elements only have a start tag; end tags must not be specified for void elements. Foreign elements must either have a start tag and an end tag, or a start tag that is marked as self-closing, in which case they must not have an end tag.

This means that `<div/>` is invalid, while `<path/>` (which is a SVG element) is OK. This configuration option would allow users to take advantage of the terser SVG syntax, while still using correct HTML syntax for normal elements (like `div`).

Closes #837.


packages/eslint-plugin-svelte/src/utils/element-types.ts
```diff
@@ -0,0 +1,140 @@
+export const voidElements = [
```
jrmajor, line 1: Moved from removed `void-elemets.ts` file. It's actually different that the list that Svelte uses (https://github.com/sveltejs/svelte/blob/c32a91891fc647998f738371a012fcba97d56e50/packages/svelte/src/utils.js#L16-L33).

packages/eslint-plugin-svelte/src/utils/element-types.ts
```diff
@@ -0,0 +1,140 @@
...
+	'link',
+	'menuitem',
+	'meta',
+	'param',
+	'source',
+	'track',
+	'wbr'
+];
+
+export const svgElements = [
```
jrmajor, line 20: Copied from https://github.com/sveltejs/svelte/blob/c32a91891fc647998f738371a012fcba97d56e50/packages/svelte/src/utils.js#L262-L349.

packages/eslint-plugin-svelte/src/utils/element-types.ts
```diff
@@ -0,0 +1,140 @@
...
+	'textPath',
+	'tref',
+	'tspan',
+	'unknown',
+	'use',
+	'view',
+	'vkern'
+];
+
+export const mathmlElements = [
```
jrmajor, line 109: Copied from https://github.com/sveltejs/svelte/blob/c32a91891fc647998f738371a012fcba97d56e50/packages/svelte/src/utils.js#L356-L387.

ota-meshi (approved):
Thank you for this PR! LGTM!

PR #842, opened 2024-08-18
[doc] Update README and user guide to use eslint flat format for TS configuration

xiaohk:
This PR updates the TS sections in README and `user-guide.md` by replacing the legacy `eslintrc.*` configuration with the eslint flat config `eslint.config.js`.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -115,56 +115,57 @@ module.exports = {
 };
 ```
 
-For example, if you are using the `"@typescript-eslint/parser"`, and if you want to use TypeScript in `<script>` of `.svelte`, you need to add more `parserOptions` configuration.
+For example, if you are using the `"typescript-eslint/parser"`, and if you want to use TypeScript in `<script>` of `.svelte`, you need to add more `parserOptions` configuration.
 
 ```js
-module.exports = {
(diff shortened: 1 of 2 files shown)


docs/user-guide.md
```diff
@@ -115,56 +115,57 @@ module.exports = {
...
-    extraFileExtensions: ['.svelte'] // This is a required setting in `@typescript-eslint/parser` v4.24.0.
+import svelteParser from 'svelte-eslint-parser';
+import tseslint from 'typescript-eslint';
+
+export default [
+  //...
+  {
+    files: ['**/*.ts', '**/*.js', '**/*.svelte', '**/*.cjs'],
+    languageOptions: {
+      parser: tseslint.parser,
```
ota-meshi, line 129: I don't think we need to use `tseslint.parser` for `*.js` and `*.svelte`, I think it makes sense to specify only `parserOptions` here.

```suggestion
```

Maybe we need another section that only specifies `parser: tseslint.parser`. What do you think?

ota-meshi (changes requested):

ota-meshi:
I'm closing this PR as there has been no progress for a long time.

PR #844, opened 2024-08-26
feat: Add fixes for block-lang rule

Winter:
I wanted eslint to automatically fix script/style errors from the block-lang rule.

The `lang` fixes will only apply when a script/style config value contains only a single, non-null value.

```html
<!-- svelte/block-lang: ["error", {"enforceScriptPresent": true}] -->
<!-- Error: The <script> block should be present and its lang attribute should be omitted -->

<!-- fixes into at the top of the file --> 
<script>
</script>
```

```html
<!-- svelte/block-lang: ["error", {"enforceScriptPresent": true, "script": "ts"}] -->
<!-- Error: The <script> block should be present and its lang attribute should be "ts" -->

<!-- fixes into at the top of the file --> 
<script lang="ts">
</script>
```

```html
<!-- svelte/block-lang: ["error", {"script": "ts"}] -->
<!-- Error: The lang attribute of the <script> block should be "ts"-->

<script>
</script>
<!-- fixes into --> 
<script lang="ts">
</script>
```

```html
<!-- svelte/block-lang: ["error", {"script": "ts"}] -->
<!-- Error: The lang attribute of the <script> block should be "ts"-->

<script context="module">
</script>
<!-- fixes into --> 
<script lang="ts" context="module">
</script>
```

```html
<!-- svelte/block-lang: ["error", {"script": "ts"}] -->
<!-- Error: The lang attribute of the <script> block should be "ts" -->

<script module>
</script>
<!-- fixes into --> 
<script lang="ts" module>
</script>
```
*The same logic applies to the `style` tag except will create at the bottom of the file*


ota-meshi (changes requested):
Thank you for this PR!

I think changing the `lang` attribute may break the component.
Auto-fix needs to be safe, so could you change it to use the suggestions API instead?
If multiple langs ​​are allowed, we could show multiple suggestions.

packages/eslint-plugin-svelte/tests/fixtures/rules/block-lang/invalid/script/javascript/javascript-as-style-lang01-errors.yaml
```diff
@@ -1,4 +1,9 @@
 - message: The lang attribute of the <script> block should be "javascript".
   line: 1
   column: 1
-  suggestions: null
+  suggestions:
+    - desc: Add a <script> block with the lang attribute set to "javascript".
+      output: |
+        <script lang="javascript"></script>
+
+        <style lang="javascript"></style>
```
baseballyama, line 9: This is wrong. I will fix this but I need to go outside now.

baseballyama: Upon carefully reviewing the test cases, this is correct.😇

baseballyama (approved):

PR #856, opened 2024-09-15
chore: Avoid using deprecated FlatConfig eslint type

KuSh:
This type has been deprecated in eslint, see https://github.com/eslint/eslint/blob/71f37c5bf04afb704232d312cc6c72c957d1c14e/lib/types/index.d.ts#L1306

### DIFF
--- a/.changeset/warm-eyes-boil.md
+++ b/.changeset/warm-eyes-boil.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': patch
+---
+
+chore: Avoid using deprecated FlatConfig eslint type
--- a/docs-svelte-kit/eslint.config.mjs
+++ b/docs-svelte-kit/eslint.config.mjs
@@ -2,7 +2,7 @@ import * as myPlugin from '@ota-meshi/eslint-plugin';
 import globals from 'globals';
 
 /**
- * @type {import('eslint').Linter.FlatConfig[]}
+ * @type {import('eslint').Linter.Config[]}
  */
 const config = [
 	{
--- a/docs-svelte-kit/src/lib/eslint/scripts/linter.js
+++ b/docs-svelte-kit/src/lib/eslint/scripts/linter.js
@@ -125,7 +125,7 @@ export function getRule(ruleId) {
 }
 
 /**
- * @returns {import('eslint').Linter.FlatConfig[]}
+ * @returns {import('eslint').Linter.Config[]}
  */
 export function createLinterConfig() {
 	return [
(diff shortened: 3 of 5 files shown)


ota-meshi (commented):
Thank you for this PR!
But our plugin can still be used with ESLint v8. Maybe this change will cause type errors for users who use ESLint v8.
So I'm worried that this should be changed in the next major version of eslint-plugin-svelte. What do you think?

KuSh:
> Maybe this change will cause type errors for users who use ESLint v8. So I'm worried that this should be changed in the next major version of eslint-plugin-svelte. What do you think?

I dont think so or they already have the problem because the latest type is used everywhere exept for that specific config, see https://www.npmjs.com/package/eslint-plugin-svelte?activeTab=code

ota-meshi:
Are you saying you don't think using ESLint v8 would cause any problems? Have you tried that?

KuSh:
> Are you saying you don't think using ESLint v8 would cause any problems? Have you tried that?

I'm saying the type Linter.Config is already used in the same file for all flat config entries except the flat/all one so if ESLint v8 would cause problem it already does.

See the generated types for v2.44.1:
```ts
import './rule-types';
import type { RuleModule } from './types';
import * as processor from './processor';
declare const _default: {
    meta: typeof processor.meta;
    configs: {
        base: import("eslint").Linter.LegacyConfig<import("eslint").Linter.RulesRecord, import("eslint").Linter.RulesRecord>;
        recommended: import("eslint").Linter.LegacyConfig<import("eslint").Linter.RulesRecord, import("eslint").Linter.RulesRecord>;
        prettier: import("eslint").Linter.LegacyConfig<import("eslint").Linter.RulesRecord, import("eslint").Linter.RulesRecord>;
        all: import("eslint").Linter.LegacyConfig<import("eslint").Linter.RulesRecord, import("eslint").Linter.RulesRecord>;
        'flat/base': import("eslint").Linter.Config<import("eslint").Linter.RulesRecord>[];
        'flat/recommended': import("eslint").Linter.Config<import("eslint").Linter.RulesRecord>[];
        'flat/prettier': import("eslint").Linter.Config<import("eslint").Linter.RulesRecord>[];
        'flat/all': import("eslint").Linter.FlatConfig[];
    };
    rules: {
        [key: string]: RuleModule;
    };
    processors: {
        '.svelte': typeof processor;
        svelte: typeof processor;
    };
};
export = _default;
```

ota-meshi:
In `@types/eslint` v8, `Linter.Config` was defined as a type representing the old configuration.

ota-meshi:
> See the generated types for v2.44.1:

Oh I didn't understand that. I didn't realize that eslint-plugin-svelte had already changed the types and released them.
We can assume that if no new issues are being opened then it's fine for users.
Thank you for explaining it to me.

ota-meshi (approved):
Thank you!

PR #868, opened 2024-10-10
feat(no-inspect): add `no-inspect` rule

### LINKED ISSUE #812: New rule: `svelte/no-$inspect`
### Motivation

THe motivation is same as [svelte/no-at-debug-tags](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-at-debug-tags/).the 

### Description

The `$inspect` should be removed when you no longer need it after you use it for debugging.

### Examples

```svelte
<script>
	let count = $state(0);
	let message = $state('hello');
        <!-- ✗ BAD -->
	$inspect(count, message); // will console.log when `count` or `message` change
</script>

<button onclick={() => count++}>Increment</button>
<input bind:value={message} />
```

### Additional comments

_No response_

mikededo:
Add rule that warns if `$inspect` is used.

Closes #812


packages/eslint-plugin-svelte/src/rules/no-inspect.ts
```diff
@@ -0,0 +1,30 @@
+import type { TSESTree } from '@typescript-eslint/types';
+
+import { createRule } from '../utils';
+
+export default createRule('no-inspect', {
+	meta: {
+		docs: {
+			description: '',
```
ota-meshi, line 8: Could you please add a description?

packages/eslint-plugin-svelte/src/rules/no-inspect.ts
```diff
@@ -0,0 +1,30 @@
...
+			category: 'Best Practices',
+			recommended: true,
```
ota-meshi, line 10: Could you set it to `false` since it would be a breaking change? Since we want to change it in a major version, could you leave a comment `// TODO Switch to recommended in the major version.` ?

ota-meshi (changes requested):
Thank you for this PR!

packages/eslint-plugin-svelte/src/rules/no-inspect.ts
mikededo: Yes, my bad 🤦🏼

ota-meshi (approved):
LGTM! Thank you!

PR #869, opened 2024-10-10
fix: correct new rule docs path

mikededo:
Current `new-rule.ts` script generates a doc inside the
`./packages/eslint-plugin-svelte`, instead of creating a doc in the actual
`./docs` folder.


ota-meshi (approved):
Thank you!

PR #885, opened 2024-10-20
feat!: drop support for old node versions (<18, 19, 21)

ota-meshi:


baseballyama:
Do we want to drop ESLint v8 also?😀 (At least v7?)

ota-meshi:
I would like to change to only supporting ESLint latest and v8.57.1. I want to keep v8.57.1 ​​to make it easier for users to migrate, but I'm not that strict about it. I'm okay to drop support v8.57.1.

baseballyama:
Question: Do we need to drop ESLint v8.57.1 if we want ESM only?

ota-meshi:
ESLint v8.57.1 ​​also has an (experimental) Flat Config available, so I don't think it will block us from changing it to ESM.

baseballyama:
OK! So I agreed that we will support only ESLint v8.57.1 and v9 in next major version.

baseballyama (approved):

PR #899, opened 2024-10-31
chore: Deprecated rules that are irrelevant is Svelte 5

### LINKED ISSUE #894: Rule deprecation request
Hi, I'm starting to update my projects to Svelte v5 and noticed that some of the rules I added were no longer relevant or outright incorrect. While at it, I went over all rules and I propose to deprecate the following ones (and eventually remove them once Svelte does so):

- [ ] `svelte/no-dupe-on-directives` - duplicate directives now cause an error in Svelte itself, no need for a lint rule.
- [ ] `svelte/no-dynamic-slot-name` - slots are deprecated and snippets can't have a dynamic name afaik
- [ ] `svelte/no-reactive-reassign` - Reactive statements are deprecated
- [ ] `svelte/no-immutable-reactive-statements` - Reactive statements are deprecated
- [ ] `svelte/no-reactive-functions` - Reactive statements are deprecated
- [ ] `svelte/no-reactive-literals` - Reactive statements are deprecated
- [ ] `svelte/require-event-dispatcher-types` - Events are deprecated
- [ ] `svelte/experimental-require-slot-types` - Slots are deprecated (https://github.com/sveltejs/eslint-plugin-svelte/pull/1005)
- [ ] `svelte/experimental-require-strict-events` - Events are deprecated and the `strictEvents` attribute is removed outright in Svelte v5
- [ ] `svelte/no-not-function-handler` - Callbacks are regular props in Svelte 5 and thus are checked by TS.

marekdedic:
Closes #894

### DIFF
--- a/docs/rules/experimental-require-slot-types.md
+++ b/docs/rules/experimental-require-slot-types.md
@@ -10,6 +10,8 @@ since: 'v2.18.0'
 
 > require slot type declaration using the `$$Slots` interface
 
+- :warning: This rule was **deprecated**. Slots are deprecated in Svelte 5
+
 ## :book: Rule Details
 
 This rule enforces the presence of the `$$Slots` interface if any slots are present in the component. This interface declares all of the used slots and their props and enables typechecking both in the component itself as well as all components that include it.
--- a/.changeset/fresh-lemons-rush.md
+++ b/.changeset/fresh-lemons-rush.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': major
+---
+
+chore: Deprecated rules that are irrelevant is Svelte 5
--- a/README.md
+++ b/README.md
@@ -379,17 +379,13 @@ These rules relate to possible syntax or logic errors in Svelte code:
 
 | Rule ID | Description |    |
 |:--------|:------------|:---|
-| [svelte/infinite-reactive-loop](https://sveltejs.github.io/eslint-plugin-svelte/rules/infinite-reactive-loop/) | Svelte runtime prevents calling the same reactive statement twice in a microtask. But between different microtask, it doesn't prevent. |  |
 | [svelte/no-dom-manipulating](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-dom-manipulating/) | disallow DOM manipulating |  |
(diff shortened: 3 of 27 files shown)


docs/rules/experimental-require-slot-types.md
```diff
@@ -10,6 +10,8 @@ since: 'v2.18.0'
 
 > require slot type declaration using the `$$Slots` interface
 
+- :warning: This rule was **deprecated**. Slots are deprecated in Svelte 5
```
baseballyama, line 13: I don't think this should be deprecated. Because still Svelte5 can use `slot`. And still there are a lot of Svelte3/4 users.
My suggestion is that just remove this from recommended rules config.
(Same for other rules also.)

marekdedic: Hmm, then we didn't understand each other in the original issue. Out of the discussed rules, only one was present in the recommended config - `no-dynamic-slot-name`. I proposed to mark as deprecated all rules that rely on features that are deprecated in Svelte itself. They would still be present for users of Svelte 4 and users relying on legacy svelte features. Only when Svelte itself removes the deprecated features do I think this plugin should remove the rules deprecated in this PR...

baseballyama: Sorry for the late reply🙏
Indeed, that would be true if all Svelte users were using Svelte 5, but many are still on Svelte 3 or 4. Therefore, eslint-plugin-svelte should continue supporting Svelte 3/4 as well. we could deprecate or remove Svelte 3/4 rules in version 4 of eslint-plugin-svelte.
So, my opinion is not to deprecate these rules but to remove the Svelte 3/4-specific rules from the recommended config.

ota-meshi: After hearing your opinion, I think that's a good idea 👍 
Additionally, it might be good to add documentation for those rules to say that they are no longer useful in Svelte v5.

marekdedic: Ok, let's clarify this a bit -> #907

marekdedic:
Based on the discussion here and in #907 I am closing this in favor of #943

PR #900, opened 2024-11-01
Generalized `no-goto-without-base` into `no-navigation-without-base`

### LINKED ISSUE #891: Add rule `no-link-without-base`
### Motivation

The motivation is basically similar to #675 - when doing *internal* navigation, you almost always want to prepend the base path. However, unlike `goto`, links can also be used for external navigation. To navigate this complicated issue, I propose to only check relative links (at least by default).

### Description

Add a rule that would trigger on any link without a base path. However, there are quite some edge cases

### Examples

```svelte
<script>
  import { base } from "$app/paths";
  import { base as whatever } from "$app/paths";
  import { goto } from "$app/navigation";
</script>

  <!-- ✓ GOOD -->
<a href={base + "/foo"}>Text</a>
<a href={`${base}/foo`}>Text</a>
<a href={whatever + "/foo"}>Text</a>
<a href={`${whatever}/foo`}>Text</a>
<a href="https://absolute.url">Text</a>

  <!-- ✗ BAD -->
<a href={"/foo"}>Text</a>
<a href={"/foo" + base}>Text</a>
<a href={`foo/${base}`}>Text</a>
</svelte>
```

### Additional comments

Based on https://github.com/sveltejs/eslint-plugin-svelte/pull/679#issuecomment-1975067137

marekdedic:
Closes #891


packages/eslint-plugin-svelte/src/rules/no-navigation-without-base.ts
```diff
@@ -0,0 +1,331 @@
...
+	const set = new Set<TSESTree.Identifier>();
+	for (const { node } of referenceTracker.iterateEsmReferences({
+		'$app/paths': {
+			[ReferenceTracker.ESM]: true,
+			base: {
+				[ReferenceTracker.READ]: true
+			}
+		}
+	})) {
+		const variable = findVariable(context, (node as TSESTree.ImportSpecifier).local);
```
ota-meshi, line 130: Could you please add a test case to see if the following import works?

```js
	import *  as paths from '$app/paths';
    paths.base;
```

ota-meshi (changes requested):
Wow! Thank you for implementing the useful rule!

packages/eslint-plugin-svelte/src/rules/no-navigation-without-base.ts
marekdedic: Great tip, found an issue with the implementation as well

ota-meshi:
Rule implementation looks good, but I made this package ESM.
https://github.com/sveltejs/eslint-plugin-svelte/pull/932

Could you please merge the main branch and fix it by adding `.js` to the import statement of your implementation?

baseballyama (approved):

PR #911, opened 2024-11-10
feat(no-inline-styles): allowing transitions by default

### LINKED ISSUE #902: `no-inline-styles`: Change the default value for `allowTransitions` to true
The `no-inline-styles` rule has an option controlling whether it reports the use of transitions as well. This was largely due to the fact that the built-in Svelte 4 transitions used inline styles and thus using them wasn't "safe" in a no-inline-styles webpage.

With Svelte 5, transitions [apparently](https://github.com/sveltejs/svelte/issues/6662#issuecomment-2452551888) no longer use inline styles. Thus, I'd like to change the default behaviour of the rule to allowing transitions (I think we can keep the option to disallow them for those who want it.)

marekdedic:
Closes #902


ota-meshi (approved):
Thank you!

PR #918, opened 2024-11-15
feat: add rule `no-deprecated-raw-special-elements`

### LINKED ISSUE #913: Detect invalid HTML elements
### Motivation

I found that in Svelte 4.0, we had used the following syntax:
```svelte
<head>
  <script>...</script>
</head>

<svelte:head>
   <title>...</title>
</svelte:head>

<div>
</div>
```

But it turns out that in Svelte 5.0 (and officially even before Svelte 5.0, even though it worked :P) this is not valid syntax.

### Description

The idea is to have a list of Svelte tags (known as "Special elements" in the Svelte docs) which are only supported with the `svelte:` syntax, e.g. `head`, `body`, `window`, `document`, `element` and to automatically notify the user if they are incorrect using these tags without the `svelte:` prefix. AFAIK, the linter already blocks components from using lower-case names (`component-name-lowercase`), so I believe that this rule would be exceedingly unlikely to trigger false positives.

### Examples

```svelte
<script>
</script>

<!-- ✓ GOOD -->
<svelte:head>
...
</svelte:head>

<!-- ✗ BAD -->
<head>
...
</head>
```

### Additional comments

_No response_

mikededo:
~Adds a new Svelte 5 specific rule that ensure special elements are used with
`svelte:` prefix. This rule will help on migrating from Svelte 4, as such
elements were supported without the prefix in Svelte 4.~

Adds a new rule that recommends not using raw special elements and fixes to
using the `svelte:` prefix. Raw special elements are deprecated from v5 on.

Closes #913


mikededo:
Hey @ota-meshi, would it be possible to have this rule have different test outputs for each svelte version? As fixes only apply to Svelte 5

ota-meshi:
You can exclude tests for each package version you are testing by placing a `_requirements.json` in your test fixtures directory and listing the versions your tests require.
However, I don't think you need to do this if your tests pass.

mikededo:
The rule should only be tested in Svelte 5, meaning tests for earlier versions should not produce errors related to Svelte 5 behavior.

packages/eslint-plugin-svelte/src/rules/no-invalid-html-elements.ts
```diff
@@ -0,0 +1,54 @@
...
+		messages: {
+			invalidElement: 'Invalid {{name}} element, use svelte:{{name}} instead.'
+		},
+		type: 'problem', // 'problem', or 'layout',
+		fixable: 'code'
+	},
+	create(context) {
+		const sourceCode = getSourceCode(context);
+		const ctx = sourceCode.parserServices.svelteParseContext;
+		if (!(ctx?.compilerVersion ?? '').startsWith('5')) {
```
ota-meshi, line 26: I don't think this branch is necessary in my opinion. `<svelte:*>` is available in v4 too. I think it makes sense for people to use the new way even if they're using v4. What do you think?

mikededo: Yes, that's true, the only thing is that the elements are _valid_ in Svelte 4, for example. That's why I preferred to report an error only in Svelte 5 or higher, where such elements are not valid.

ota-meshi: Hmm. How about renaming the rule to encourage its use even in Svelte v4? How about naming it `no-deprecated-raw-special-elements`?

mikededo: And would we report it as an error for v4, or something more as a warning?

ota-meshi: For now, I think it's user preference.
I'm not sure yet how we'll be warned if we add new rules to the sharable config.
If we add a sharable configuration for v4, we might be able to warn when we use it.

ota-meshi (approved):
LGTM! Thank you!

baseballyama:
@mikededo 

The rule is named "no-deprecated-raw-special-elements" but where is it stated that these elements are actually "deprecated"? For example, using `<head>` in a Svelte component is almost certainly incorrect even in Svelte 4. I believe this rule remains valid in Svelte4 also.

mikededo:
@baseballyama would it make more sense to replace the `deprecated` for `invalid`?

baseballyama:
@mikededo 

Whether it’s invalid or not depends on how users use it (although in almost all cases, it would indeed be invalid😅). A name like `no-raw-special-elements` might be more appropriate. Additionally, while the documentation states that this is specific to Svelte 5, I believe this rule could also be applied to Svelte 3 and 4.

mikededo:
@baseballyama True. It was not valid in previous versions but contrary to Svelte 5, the elements did work. 

I can work on updating it.

baseballyama:
Thank you so much!

mikededo:
@baseballyama fixed in #1015!

PR #925, opened 2024-11-22
feat: added the `consistent-selector-style` rule

### LINKED ISSUE #898: Add rule `consistent-selector-style`
### Motivation

When styling elements in Svelte components, there are many options for referencing said elements. The simplest one, leading to the least amount of code is to simply refer to them by their tag name. I'd like to add a rule that enforces this if it is possible (i.e. if there are no duplicates of said element). This may dovetail nicely with `no-unused-class-name` and lead to code reduction

### Description

For each CSS selector using a class, the rule would check the tag name for the element the class is referring to. If there aren't multiple elements with that same tag (or they all share the same class), the rule would trigger and recommend selecting by the tag name instead. The same would apply to IDs

### Examples

```svelte
<a class="link">Text 1</a>

<b class="bold">Text 2</b>
<b class="bold">Text 3</b>

<i class="italic">Text4</i>
<i>Text5</i>

<style>
  /* ✗ BAD */
  .link {}
  .bold {}

  /* ✓ GOOD */
  .italic {}
</style>
```

### Additional comments

_No response_

marekdedic:
Closes #898

### DIFF
--- a/docs/rules/consistent-selector-style.md
+++ b/docs/rules/consistent-selector-style.md
@@ -1,0 +1,51 @@
+---
+pageClass: 'rule-details'
+sidebarDepth: 0
+title: 'svelte/consistent-selector-style'
+description: 'enforce a consistent style for CSS selectors'
+---
+
+# svelte/consistent-selector-style
+
+> enforce a consistent style for CSS selectors
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> **_This rule has not been released yet._** </badge>
+
+## :book: Rule Details
+
+This rule reports ???.
+
+<ESLintCodeBlock>
+
+<!--eslint-skip-->
+
+```svelte
+<script>
+  /* eslint svelte/consistent-selector-style: "error" */
+</script>
+
+<!-- ✓ GOOD -->
+
+<!-- ✗ BAD -->
+```
+
+</ESLintCodeBlock>
+
+## :wrench: Options
+
+```json
+{
+  "svelte/consistent-selector-style": ["error", {}]
+}
+```
+
+-
+
+## :books: Further Reading
+
+-
+
+## :mag: Implementation
+
+- [Rule source](https://github.com/sveltejs/eslint-plugin-svelte/blob/main/packages/eslint-plugin-svelte/src/rules/consistent-selector-style.ts)
+- [Test source](https://github.com/sveltejs/eslint-plugin-svelte/blob/main/packages/eslint-plugin-svelte/tests/src/rules/consistent-selector-style.ts)
--- a/docs/rules/consistent-selector-style.md
+++ b/docs/rules/consistent-selector-style.md
@@ -13,37 +13,79 @@ description: 'enforce a consistent style for CSS selectors'
 
 ## :book: Rule Details
 
-This rule reports ???.
-
-<ESLintCodeBlock>
+This rule allows you to set a preferred style for your CSS (& other style language) selectors. In CSS, there is a wide list of options for selecting elements, however, the three most basic types select by element type (i.e. tag name), ID or class. This rule allows you to set a preference for some of these three styles over others. Not all selectors can be used in all situations, however. While class selectors can be used in any situation, ID selectors can only be used to select a single element and type selectors are only applicable when the list of selected elements is the list of all elements of the particular type. To help with this, the rule accepts a list of selector style preferences and reports situations when the given selector can be rewritten using a more preferred style.
 
 <!--eslint-skip-->
 
 ```svelte
 <script>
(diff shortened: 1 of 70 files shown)


marekdedic:
Hi, @ota-meshi! This one is I think ready for review, however, it needs sveltejs/svelte-eslint-parser#619 and #965 first.

packages/eslint-plugin-svelte/src/rules/consistent-selector-style.ts
```diff
@@ -0,0 +1,286 @@
...
+
+/**
+ * Checks whether a given selection could be obtained using a type selector
+ */
+function couldBeType(
+	selection: AST.SvelteHTMLElement[],
+	typeSelections: Map<string, AST.SvelteHTMLElement[]>
+): boolean {
+	const types = new Set(selection.map((node) => node.name.name));
+	if (types.size > 1) {
```
marekdedic, line 257: This doesn't report any selection which could be replaced by multiple type selections (e.g. replacing `.emphasized {}` by `i, b {}`). That can be added later I think (maybe gated with a configuration option?)

packages/eslint-plugin-svelte/src/rules/consistent-selector-style.ts
```diff
@@ -0,0 +1,292 @@
...
+/**
+ * Compares two arrays for item equality
+ */
+function arrayEquals(array1: AST.SvelteHTMLElement[], array2: AST.SvelteHTMLElement[]): boolean {
+	function comparator(a: AST.SvelteHTMLElement, b: AST.SvelteHTMLElement): number {
+		return a.range[0] - b.range[0];
+	}
+
+	const array2Sorted = array2.slice().sort(comparator);
+	return (
```
ota-meshi, line 283: Couldn't you compare it like this?

```js
return (
  array1.length === array2.length && 
  array1.every(e => array2.includes(e))
)
```

ota-meshi (changes requested):
Thank you for this PR!
I have a few comments.

packages/eslint-plugin-svelte/src/rules/consistent-selector-style.ts
marekdedic: You could, but it's slower (n² instead of n log n)

packages/eslint-plugin-svelte/src/rules/consistent-selector-style.ts
```diff
@@ -0,0 +1,292 @@
...
+	key: string,
+	value: AST.SvelteHTMLElement
+): void {
+	map.set(key, (map.get(key) ?? []).concat(value));
+}
+
+/**
+ * Checks selectors in a given PostCSS node
+ */
+function checkSelectorsInPostCSSNode(node: AnyNode, ruleGlobals: RuleGlobals): void {
```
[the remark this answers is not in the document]
marekdedic: You could, but `ruleGlobals` need to be passed down so many functions... Putting them all in `create` would make it really long and unreadable, so I don't think that's a good idea, what do you think?

docs/rules/consistent-selector-style.md
```diff
@@ -0,0 +1,93 @@
...
+    {
+      "checkGlobal": false,
+      "style": ["type", "id", "class"]
+    }
+  ]
+}
+```
+
+- `checkGlobal` ... Whether to check styles in `:global` blocks as well. Default `false`.
+- `style` ... A list of style preferences. Default `["type", "id", "class"]`.
```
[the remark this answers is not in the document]
marekdedic: I added a better description further up the file in the rule details.

packages/eslint-plugin-svelte/src/rules/consistent-selector-style.ts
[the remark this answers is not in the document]
ota-meshi: I think I understand your point.

However, in my opinion, I would rather have a longer `create()` function, which would make it harder to read, than having to pass many properties to the current `ruleGlobals` and make them harder to keep track of.
I think that RuleGlobals can't be understood without reading the `create()` function after all.

But, I may not always be of this opinion. I may have a different opinion for a different rule implementation, but that's how I feel about this PR.

packages/eslint-plugin-svelte/src/rules/consistent-selector-style.ts
ota-meshi: In my opinion, I don't expect any performance issues with non-malicious `.svelte` files (please let me know if this isn't the case).
For now I prefer to keep the function size at 2 lines.

ota-meshi (approved):
LGTM! Thank you!

PR #932, opened 2024-11-28
feat!: change the plugin to an ESM-only package

### LINKED ISSUE #808: ESM-only plugin
### Description

You can now write eslint plugins in ESM. Their docs demonstrate it here: https://eslint.org/docs/latest/extend/plugins

This would be a nice code modernization. It would also potentially be helpful to the Svelte core project as one the big reasons we've had to distribute a CJS version of the Svelte compiler is to support the eslint plugin. Though the other reason we need it is for the REPL and I'm unsure if we can avoid it there.

I think this may only work in eslint 9, so the plugin would need to drop support for older versions.

ota-meshi:
This PR changes the plugin to an ESM-only package.
~~However, it is not yet a fully ESM-only package, as we have not yet made the parser ESM.~~

Close #808


baseballyama:
So, after upgrade parser version to `[v1.0.0-next.2](https://github.com/sveltejs/svelte-eslint-parser/releases/tag/v1.0.0-next.2)`, it will be fully ESM-only package?

ota-meshi:
Yeah! If we upgrade the parser it becomes a complete ESM package. I'll do that at lunchtime.

ota-meshi:
I updated the parser and it seems to work fine 😄

baseballyama (approved):
I will use to test this version after release this PR.

Issue #818, opened 2024-07-06
New Rules: rune-prefer-let & prefer-const


bfanger:
### Motivation

`const` in javascript:

> The const declaration creates an immutable reference to a value. 
> It does not mean the value it holds is immutable — just that the variable identifier cannot be reassigned. 
> You should understand const declarations as "create a variable whose identity remains constant", not "whose value remains constant" — or, "create immutable bindings", not "immutable values".
>
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const

At the moment the Svelte 5 compiler [breaks this rule](https://github.com/sveltejs/svelte/issues/12121) and we can no longer make assumptions of what should be possible.
```svelte
<script>
const { value } = $props() 

const copy = value;

console.log(value) // output: 1

function onclick() {
    console.log(value) // output: 3 (prop was updated)
    if (copy === value) { // false ?!?
    }
}
</script>

<button {onclick}>Click me</button>
```
Because the compiler replaces all usages of reactive values by getters & setters, this doesn't result in a`TypeError: Assignment to constant variable.` error like it would in regular javascript.

### Description

The svelte/rune-prefer-let rule is aware of which variables are reactive by detecting switch are assigned via a rune.

eslint `prefer-const` rule doesn't see any reassignments so it assumes a let statement can be safely converted to `const` which conflicts with the svelte/rune-prefer-let

The svelte/prefer-const rule is identical, but takes into account which variable are reactive an can be reassigned.

### Examples

```svelte

<!-- ✓ GOOD -->
<script>
let {count,  onclick} = $props();

let double = $derived(count * 2);
</script>


<!-- ✗ BAD -->
<script>
const {count,  onclick} = $props();

const double = $derived(count * 2);
</script>

```

### Additional comments

Implementation of these rules are available https://github.com/sveltejs/eslint-plugin-svelte/pull/806 & https://github.com/sveltejs/eslint-plugin-svelte/pull/816

ota-meshi:
The code doesn't change the variables, so I think it's fine to use `const`. Is it really necessary to use `let`? :thinking: 
They work in runes, so it makes sense that they don't match the description on MDN.

mrh1997:
@ota-meshi : Although I absolutely can follow your argumentation please keep in mind that this is not what people will intuitively do.

Thus if you stick to your approach all tutorials should be adopted and eslint should at least output a message which tells the user to use "const" when working with "$derived".

[...]

PR #933, opened 2024-11-28
feat(prefer-const): add rule

mikededo:
Adds a new rule named `prefer-const` which:

- Overrides the ESLint `prefer-const rule to support Svelte syntax, as discussed
  in #818.
- ~Ports~ Proxies the original `prefer-const` rule from ESLint, adapting it for Svelte.

~This PR is an updated version of #816, which did not include types. Also, some
opinionated changes were made in terms of code style, not in terms of behavior.~
Avoided re-implemeting the core rule, by using the `getCoreRule` and proxying the
default implementation.

I found that this rule could be very useful for Svelte, and there could be many
developers who would benefit from it. ~At the same time, I thought that the
initial implementation was a good starting point, and that adding types and
would make the rule more robust to changes.~  
However, the rule should not have many changes, as how variables are handled in
JS is not likely to change.

> Acknowledgement: this rule is based on the `prefer-const` rule from ESLint and
> the initial work done by @bfanger.


ota-meshi:
Thanks for working on implementing this rule!
However, could you please implement it using `getCoreRule()` in the same way as `no-inner-declarations`?
https://github.com/sveltejs/eslint-plugin-svelte/blob/main/packages/eslint-plugin-svelte/src/rules/no-inner-declarations.ts

I think it just avoids calling listeners on `VariableDeclaration` that use runes.

mikededo:
Thanks @ota-meshi for pointing this out, my initial idea was to use a proxy for the rule, to avoid having a complete re-write. Should've asked before spending so many hours 😅 🤦🏼 

Will look into it. Maybe I'll end up closing this PR in favour of another one.

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
```diff
@@ -0,0 +1,99 @@
...
+import { defineWrapperListener, getCoreRule } from '../utils/eslint-core';
+
+const coreRule = getCoreRule('prefer-const');
+
+type Declaration = TSESTree.Expression;
+
+/**
+ * Finds and returns the callee of a declaration node within variable declarations or object patterns.
+ */
+function findDeclarationCallee(node: Declaration) {
```
ota-meshi, line 13: I don't think we need a type alias.

```suggestion
function findDeclarationCallee(node: TSESTree.Expression) {
```

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return null;
+}
+
+/**
+ * Determines if a declaration should be skipped in the const preference analysis.
+ * Specifically checks for Svelte's state management utilities ($state, $props, $derived).
+ */
+function shouldSkipDeclaration(declaration: Declaration | null) {
```
ota-meshi, line 35:
```suggestion
function shouldSkipDeclaration(declaration: TSESTree.Expression | null) {
```

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Finds and returns the callee of a declaration node within variable declarations or object patterns.
+ */
+function findDeclarationCallee(node: Declaration) {
+	const { parent } = node;
+	if (parent.type === 'VariableDeclarator' && parent.init?.type === 'CallExpression') {
+		return parent.init.callee;
+	}
+
+	if (
```
ota-meshi, line 19: What code passes through this branch?

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
```diff
@@ -0,0 +1,99 @@
...
+	if (!declaration || !declaration.parent) {
+		return false;
+	}
+
+	const callee = findDeclarationCallee(declaration);
+	if (!callee) {
+		return false;
+	}
+
+	if (callee.type === 'Identifier' && ['$state', '$props', '$derived'].includes(callee.name)) {
```
ota-meshi, line 45: Why did you exclude `$state` from rule checking?
I think it's ok to have a `const` variable as long as the user doesn't reassign it in their code 🤔

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
```diff
@@ -0,0 +1,99 @@
...
+		callee.object.name === '$derived' &&
+		callee.property.type === 'Identifier' &&
+		callee.property.name === 'by'
+	) {
+		return true;
+	}
+	if (
+		callee.object.name === '$state' &&
+		callee.property.type === 'Identifier' &&
+		callee.property.name === 'frozen'
```
ota-meshi, line 63: I think it's `$state.raw`.

```suggestion
		callee.property.name === 'raw'
```

ota-meshi (changes requested):
Thank you for this PR!

packages/eslint-plugin-svelte/package.json
```diff
@@ -30,7 +30,7 @@
     "cover": "nyc --reporter=lcov pnpm run test",
     "debug": "pnpm run mocha \"tests/src/**/*.ts\" --reporter dot --timeout 60000",
     "lint": "run-p lint:*",
-    "lint-fix": "pnpm run lint:es --fix && pnpm run lint:style --fix",
+    "lint:fix": "pnpm run lint:es --fix",
```
ota-meshi, line 33: Can you revert the changes?

ota-meshi: Oh, `lint:style` doesn't exist now. Could you revert the name of script?

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
mikededo: This was part of the discussion in #818. I also proposed having an option that allows the user to decide [which reactive statements](https://github.com/sveltejs/eslint-plugin-svelte/issues/818#issuecomment-2505770979) should be excluded from this rule.

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
mikededo: After investigating for a bit, I realised that there's no valid JavaScript/TypeScript syntax that would create an AST structure, taking into account that the node that's passed into `findDeclarationCallee` comes from a `VariableDeclaration.declarations[number].init`.  The structure doesn't make sense because a `Property` node’s parent cannot simultaneously be a `VariableDeclarator`. 

I'll delete this branch.

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
ota-meshi: I read the thread. So it's ok for you to use `const` for `$state` without reassignment, right?
I'm also ok with `const`, so I think it's better to remove the `$state` branch from the PR and check `$state`.

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
```diff
@@ -0,0 +1,88 @@
+import { createRule } from '../utils';
+import type { TSESTree } from '@typescript-eslint/types';
+
+import { defineWrapperListener, getCoreRule } from '../utils/eslint-core';
```
ota-meshi, line 4: I changed this package to ESM.
https://github.com/sveltejs/eslint-plugin-svelte/pull/932

Could you please merge the main branch and fix it by adding `.js` to the import statement of this PR?

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
mikededo: Yes!

Also, what about the option to customize which reactive assignments to ignore?

ota-meshi: > Also, what about the option to customize which reactive assignments to ignore?

I don't think there's a need for an option for now. If someone needs one, I think they'll suggest it in an issue. If necessary, we'll add the option then.

mikededo:
Hey @ota-meshi, would it be possible to have another review when possible? 🎅🏼

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
```diff
@@ -0,0 +1,88 @@
...
+
+	if (
+		callee.object.name === '$derived' &&
+		callee.property.type === 'Identifier' &&
+		callee.property.name === 'by'
+	) {
+		return true;
+	}
+	if (
+		callee.object.name === '$state' &&
```
ota-meshi, line 50: I think there is probably no need to ignore `$state.raw` from the check for now.


https://github.com/sveltejs/eslint-plugin-svelte/pull/933#discussion_r1865388045

ota-meshi (changes requested):
Sorry for the late reply 🙇
It almost looks good to me, but I have one comment.

packages/eslint-plugin-svelte/src/rules/prefer-const.ts
```diff
@@ -0,0 +1,81 @@
...
+	if (parent.type === 'VariableDeclarator' && parent.init?.type === 'CallExpression') {
+		return parent.init.callee;
+	}
+
+	return null;
+}
+
+/**
+ * Determines if a declaration should be skipped in the const preference analysis.
+ * Specifically checks for Svelte's state management utilities ($state, $props, $derived).
```
baseballyama, line 22:
```suggestion
 * Specifically checks for Svelte's state management utilities ($props, $derived).
```

I think here doesn't handle `$state`.

baseballyama (approved):

Issue #824, opened 2024-07-12
Publishing docs on svelte.dev or highlighting them there


benmccann:
### Description

We're starting to put together plans to publish docs for all the various plugins directly on svelte.dev  (e.g. vite-plugin-svelte, rollup-plugin-svelte, svelte-loader, svelte-preprocess, the VS code plugin, prettier-plugin-svelte, eslint-plugin-svelte, etc.)

eslint-plugin-svelte is much different than the others because it has many more pages as well as an existing site and a playground. This raises several questions:
- Should we try to incorporate the playground into the main site? [It's setup](https://github.com/sveltejs/eslint-plugin-svelte/blob/5d53db1bda21ea7cab53e7618935dc7293f10e9e/docs-svelte-kit/vite.config.mts#L43) looks a little complicated. I also don't know if that would make things harder for you?
- Should we just have a single page with setup instructions and then link to the existing site?
- Or should move all the content over except for the playground?

I've asked these questions of the other maintainers working on the site, but thought it would be good to get your input as well. We'd like to try to have the markdown continue to live in each individual repository and then the site's build and deploy process will check out each of those repositories and use the docs from there. We're working on prototyping that right now. The main change needed here would probably be to publish a bit of config stating what order the pages should show up in the sidebar (e.g. https://github.com/sveltejs/svelte/pull/12410)

ota-meshi:
Thank you for letting me know.

> Or should move all the content over except for the playground?

I like this option :+1:

I want to show red markers in the documentation for each rule, so do you think we can introduce [twoslash-eslint] instead of rule playground?
You can see a working example of twoslash-eslint at the following site:
https://eslint.style/rules/default/array-bracket-newline#always

[twoslash-eslint]: https://twoslash.netlify.app/packages/eslint


I think [the plugin playground](https://sveltejs.github.io/eslint-plugin-svelte/playground/) could be directed to another site.

https://eslint-online-playground.netlify.app/#eslint-plugin-svelte%20with%20typescript

PR #936, opened 2024-11-28
docs: use twoslash-eslint instead of playground

ota-meshi:
Related to https://github.com/sveltejs/eslint-plugin-svelte/issues/824#issuecomment-2224221355


baseballyama (approved):
I am a little disappointed that we can no longer modify the code directly on each rule page to check it.

PR #943, opened 2024-11-30
chore: removed Svelte 4 specific rules from recommended set

marekdedic:

PR #957, opened 2024-12-14
breaking: Redesigned ESLint config

baseballyama:
Svelte 5 and Svelte 4 have different features. Additionally, Svelte 5 introduces Runes mode and legacy mode, with some features marked as deprecated.
For a new project, rules related to deprecated features and legacy mode are likely unnecessary. While the Svelte version can be automatically determined from package.json, However ESLint cannot automatically detect whether deprecated features are being used without walking the AST.
Additionally, I personally dislike longer ESLint execution times caused by unnecessary rules, so I prefer to turn off unneeded rules at the configuration level. Configuring ESLint is a rare task, so even if some adjustments are required in the next major version, I believe it’s better to prioritize faster linting in daily workflows.

To address this, this PR provides the following three configurations:
- A recommended ruleset for all features available in Svelte 5.
- A recommended ruleset for Svelte 5 without legacy/deprecated features.
- A recommended ruleset for Svelte 3/4.

Alternative of https://github.com/sveltejs/eslint-plugin-svelte/pull/943


packages/eslint-plugin-svelte/tools/update-meta.ts
```diff
@@ -1,5 +1,5 @@
 import path from 'path';
-import { name, version } from '../package.json';
+import packageJson from '../package.json' assert { type: 'json' };
```
baseballyama, line 2: To solve TypeScript error.

packages/eslint-plugin-svelte/tools/update-rule-types.ts
```diff
@@ -9,7 +9,6 @@ void main();
 async function main() {
 	const { pluginsToRulesDTS } = await import('eslint-typegen/core');
 
-	// @ts-expect-error - types are a bit strict here
```
baseballyama, line 12: I don't get TypeScript error here.

baseballyama:
I think https://github.com/sveltejs/eslint-plugin-svelte/pull/980 is better.

marekdedic:
> I think #980 is better.

These are mutually exclusive, do I understand that correctly?

baseballyama:
Yes, if we merge #980, I will close this PR.

PR #965, opened 2024-12-14
chore: using svelte-eslint-parser for style selector parsing

marekdedic:
Changes based on sveltejs/svelte-eslint-parser#619


packages/eslint-plugin-svelte/src/types.ts
```diff
@@ -207,6 +210,10 @@ export interface SourceCode {
 		isSvelteScript?: boolean;
 		getSvelteHtmlAst?: () => unknown;
 		getStyleContext?: () => StyleContext;
+		getStyleSelectorAST: (rule: StyleRule) => SelectorRoot;
```
marekdedic, line 213: Some of these should have been here before, but for some reason weren't.

baseballyama (approved):

PR #970, opened 2024-12-22
fix: prevent infinite recurstion in indent

BlankParenthesis:
The error here is tricky to track down the root cause of (especially for someone new to the codebase) but the trigger for it as I understand is as follows:

1. The indent rule begins to build `offsets` as a map of token beginnings to indent levels, each referencing its own parent. Among them are "root nodes" ­— entries in the map which have a `type` of 2 and no parent.
2. A token will arrive with a starting index which overlaps a root node.
3. The offset builder then inserts this new offset on top of the root node, providing it a new parent.
4. This parent originally depended on that root node at some point in its ancestor chain, and thus a circular dependency is created.
5. When later resolving this circular dependency, infinite recursion is occurs. This results in a stack limit error, crashing the program.

The ideal solution is probably to avoid inserting these tokens which overlap root nodes, or possibly disambiguate offsets by more than simply the start index. The fix here (simply ignoring attempts to overwrite root nodes) passes all the existing tests as far as I can see but I have not had time to extensively test it and simply discarding the information seems misguided — though I cannot currently imagine a scenario where the token which overlaps the root has a different indent level from the node that it overlaps.

The test here is far from ideal but serves to demonstrate a reproducible case of the bug. Having something to check for this sort of error is good, but someone with more knowledge of the root cause might be able to produce a better fix and related test. Of note, it seems to be the most minimal viable example I can find since the following changes prevented the error from occurring:

- Removing the store unwrap token (`$`)
- Removing the Enum
- Making the Enum empty
- Changing the script lang attribute to the empty string
- Removing the script lang attribute
- Adding `<!-- prettier-ignore -->` to the top of the file

### DIFF
--- a/packages/eslint-plugin-svelte/src/rules/indent-helpers/offset-context.ts
+++ b/packages/eslint-plugin-svelte/src/rules/indent-helpers/offset-context.ts
@@ -48,6 +48,10 @@ export class OffsetContext {
 		if (index === base) {
 			return;
 		}
+		const previousOffset = this.offsets.get(index);
+		if (previousOffset?.type === 2) {
+			return;
+		}
 		this.offsets.set(index, {
 			type: OffsetDataType.normal,
 			base,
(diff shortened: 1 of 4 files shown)


BlankParenthesis:
Some further testing has revealed that this is indeed not an appropriate fix for the underlying cause: while this prevents crashing when encountering it, the reported required indent levels are often wrong for files containing both an enum and a store dereference (`$`). I've changed this to a draft for now since this feels like the sort of issue that should be addressed directly.

AlbertMarashi:
May be related:

https://github.com/sveltejs/eslint-plugin-svelte/issues/998

Issue #438, opened 2023-04-13
New Rule: svelte/kit-parent-function-at-bottom rule


baseballyama:
### Motivation

I would like to recognize performance overhead regarding `parent`.

```js
/** @type {import('./$types').PageLoad} */
export async function load({ params, parent }) {
  const parentData = await parent();
  //                       ^ `parent` function can execute later to avoid a waterfall
  const data = await getData(params);

  return {
    ...data
    meta: { ...parentData.meta, ...data.meta }
  };
}
```

### Description

SvelteKit has `parent` function. And according to the document, `parent` should use at the bottom of `load` function as much as possible in terms of performance.

https://kit.svelte.dev/docs/load#using-parent-data

But it's too difficult to recognize by human code review, so I would like to recognize this automatically.

### Examples

```js
/** @type {import('./$types').PageLoad} */
export async function load({ params, parent }) {
  // NG
  const parentData = await parent();
  const data = await getData(params);
  // OK
  const parentData = await parent();
  // OK (Because `doSomething` needs to use the return value of `parent`.)
  const foo = await doSomething(parentData);

  return {
    ...data
    meta: { ...parentData.meta, ...data.meta }
  };
}
```

### Additional comments

I need to think about better rule name.😅

ota-meshi:
New rule sound good to me!
We need a better rule name, as you say, but I don't have any ideas 😓

devunt:
How about `svelte/avoid-waterfall-in-kit-loads`?

ota-meshi:
Hmm. I'm not familiar with English, so I'm not sure what "waterfall" means. So I'm not sure if that naming of the rule makes sense 😓. I would like someone's opinion on that name.

baseballyama:
`waterfall` is used in the kit doc. So this word itself is natural.

baseballyama:
But to protect `waterfall`, possibly we need to take care something more in the future, so I'd like to use a name that has a specific check purpose.
I mean "This is is just check the place of `parent` function. Not check everything regarding avoiding waterfall"

PR #980, opened 2024-12-30
feat: Implement util to conditionally run lint based on Svelte version and SvelteKit routes etc

baseballyama:


baseballyama:
@ota-meshi 
Could you please review the policy of the PR? If there are no issues, I believe I can handle the remaining tasks on my own.

packages/eslint-plugin-svelte/src/rules/no-export-load-in-svelte-module-in-kit-pages.ts
```diff
@@ -19,9 +19,8 @@ export default createRule('no-export-load-in-svelte-module-in-kit-pages', {
 		type: 'problem'
 	},
 	create(context) {
-		if (!isKitPageComponent(context)) {
-			return {};
-		}
+		const svelteContext = getSvelteContext(context);
+		if (svelteContext?.svelteKitFileType == null) return {};
```
baseballyama, line 23: This is example how to use `getSvelteContext`.

packages/eslint-plugin-svelte/src/utils/svelte-kit.ts
baseballyama, line 1: This file's functions are moved to `packages/eslint-plugin-svelte/src/utils/svelte-context.ts`.

packages/eslint-plugin-svelte/tests/utils/utils.ts
```diff
@@ -119,14 +119,14 @@ export function loadTestCases(
...
-	const fixable = plugin.rules[ruleName].meta.fixable != null;
+	const fixable = plugin.rules[ruleName].meta?.fixable != null;
 
 	const invalid = listupInput(invalidFixtureRoot)
 		.filter(filter)
 		.map((inputFile) => {
 			const config = getConfig(ruleName, inputFile);
-			const errorFile = inputFile.replace(/input\.[a-z]+$/u, 'errors.yaml');
-			const outputFile = inputFile.replace(/input\.[a-z]+$/u, 'output.svelte');
+			const errorFile = inputFile.replace(/(input|\+.+)\.[a-z]+$/u, 'errors.yaml');
```
baseballyama, line 128: I need to support SvelteKit route components (e.g. `+page.svelte`) as input.
So I modified the file.

packages/eslint-plugin-svelte/tools/new-rule.ts
```diff
@@ -62,7 +63,8 @@ export default createRule('${ruleId}', {
         type: 'suggestion', // 'problem', or 'layout',
     },
     create(context) {
-        
+				// TODO: Please refer to the context regarding Svelte and SvelteKit, and return early if run the rule is unnecessary.
+				const svelteContext = getSvelteContext(context);
```
baseballyama, line 67: I updated new rule file template also.

marekdedic:
Hi @baseballyama, I am looking at the three alternative proposals (this one, #957, #943) and I am a little bit torn. Implementation-wise, I like this one the most because it "just works". However, I think this one will also be the hardest to explain to the users (such as me :D) clearly, because it somewhat boils down to "This will magically disable rules, trust us". In this "understandability" regard, I think the approach with separate rulesets seems easier to grasp. However, it introduces additional complexity for users, which is probably not the best choice.

Additionally, I expect that at some point in the future, obsolete versions of Svelte will be dropped altogether, so it'd be great to design this such that the users don't need to re-do their config once again at that point.

baseballyama:
@marekdedic 
Many users are not interested in configuring ESLint in detail. They don’t want to think about things like “I need this config because I’m using Svelte 5” or “I need that config because I’m using SvelteKit.” It’s much simpler and better if everything just works by using the recommended config.

This issue can also be addressed by changing perspectives. Instead of seeing it as “magic,” we can treat it as a feature to eliminate false positives. For example, if a user is only using Svelte 5, there shouldn’t be any reports related to SvelteKit rules. However, AST-based inspections may still produce false positives. With this PR, false positives can be completely eliminated.

Similarly, enabling SvelteKit-specific rules without actually using SvelteKit can also cause issues. This PR prevents such problems as well.

I plan to completely review the rules included in the recommended config from scratch after merging this PR.

marekdedic:
> Many users are not interested in configuring ESLint in detail.

Yeah, I agree... Let me rephrase: I think this is probably the best of the 3 alternatives. At the same time, I think it will require a good explanation in the documentation so that the users that *do* want to configure eslint understand what's happening in the recommended config.

marekdedic:
Sorry, looking at it more, I have a question to the design of this feature: Why is the checking of applicability of the rule left to the rule implementation? Isn't it something that could be configured in the rule meta and checked automatically? It seems to me that that would be less error-prone (especially when 3rd party contributors such as me add new rules) and maybe could be used to indicate the support in docs as well? But maybe I am missing something, I don't know the rule meta stuff that good...

baseballyama:
Is it possible to control whether a rule is invoked by accessing its meta information at runtime? (AFAIK, there is no way)
However, such improvements might be possible.

```ts
export default createRule('rule-name', {
	meta: {
		executeCondition: {
			svelte: [5],
			isKitRule: true,
			kitRouteTypes: ["+page.svelte", "+page.js"]
		},
		...
	},
	create(context) {
		return executeIf(this.meta.executeCondition, (context) => {
			// ... implement rule ...
		});
	}
});
```

marekdedic:
> Is it possible to control whether a rule is invoked by accessing its meta information at runtime?

I think it might be possible by injecting the check here:

https://github.com/sveltejs/eslint-plugin-svelte/blob/d0e76a330cf2cce18461ffa72204c71cf2ea6069/packages/eslint-plugin-svelte/src/utils/index.ts#L19

baseballyama:
Ahh. Nice! I will update the PR tomorrow. (It’s already night in Japan, so I’m going to sleep.)

baseballyama:
@marekdedic What do you think about the current implementation? I want to add tests somehow but before that I would like to ask your opinion.

packages/eslint-plugin-svelte/src/types.ts
```diff
@@ -108,6 +109,18 @@ export interface PartialRuleMetaData {
...
 	type: 'problem' | 'suggestion' | 'layout';
+	/**
+	 * Conditions to determine whether this rule should be applied.
+	 * Multiple conditions can be specified as array, and the rule will be applied if any one of them matches (logical OR).
+	 * If not specified, the rule will be applied to all files.
+	 */
+	conditions?: {
+		svelteVersion?: string; // e.g.: '>= 5.0.0'
+		fileTypes?: SvelteContext['fileType'][];
+		runes?: SvelteContext['runes'];
```
marekdedic, line 120: Why not just `boolean`? :D

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,182 @@
...
+
+	const { compilerVersion } = svelteParseContext;
+	if (compilerVersion === undefined) {
+		return null;
+	}
+
+	const filePath = getFilename(context);
+	const svelteKitContext = getSvelteKitContext(context);
+
+	const runes = svelteParseContext.runes === true;
```
marekdedic, line 169: Why compare boolean to true?

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Check givin file is under SvelteKit project.
+ *
+ * If it runs on browser, it always returns true.
+ *
+ * @param filePath A file path.
+ * @returns
+ */
+function gteSvelteKitVersion(filePath: string): string | null {
```
marekdedic, line 123: Typo (`gte` instead of `get`)

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,182 @@
...
+	// Hack: if it runs on browser, it regards as SvelteKit project.
+	if (isRunOnBrowser) return '2.15.1';
```
marekdedic, line 125: When would the linter itself run on browser?

marekdedic: nit: run *in* browser

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,182 @@
...
+import fs from 'fs';
+import path from 'path';
+import { getPackageJson } from './get-package-json.js';
+import { getFilename, getSourceCode } from './compat.js';
+
+const isRunOnBrowser = !fs.readFileSync;
+
+export type SvelteContext = {
+	svelteVersion: string;
+	fileType: '.svelte' | '.svelte.[js|ts]';
```
marekdedic, line 11: Hmm, I don't like the design of having separate `fileType` and `svelteKitFileType`. I see 2 ways to do it better:
1. Merge them
2. Don't provide these, but go conceptually one level further - why would we want to know the file type? Maybe what we actually want to know is things like
    1. Are we in a SFC file?
    2. Are we in the server context?
    3. etc.

I think the second solution is more work (and requires quite careful design), but ultimately serves us better, because rules are fundamentally going to be enabled based on the *properties* of the current context/file, not based on its type...

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: I don't really understand it, but everything to do with the `isRunOnBrowser` variable seems fishy to me...

packages/eslint-plugin-svelte/src/utils/index.ts
```diff
@@ -16,6 +81,14 @@ export function createRule(ruleName: string, rule: PartialRuleModule): RuleModul
 				ruleName
 			}
 		},
-		create: rule.create as never
+		create(context: RuleContext) {
```
marekdedic, line 84: Great, this is exactly what I imagined, glad to see it really was this easy (ignoring all the complexity in the called functions :D)

packages/eslint-plugin-svelte/src/types.ts
```diff
@@ -108,6 +109,18 @@ export interface PartialRuleMetaData {
 	deprecated?: boolean;
 	replacedBy?: string[] | { note: string };
 	type: 'problem' | 'suggestion' | 'layout';
+	/**
+	 * Conditions to determine whether this rule should be applied.
+	 * Multiple conditions can be specified as array, and the rule will be applied if any one of them matches (logical OR).
+	 * If not specified, the rule will be applied to all files.
+	 */
+	conditions?: {
+		svelteVersion?: string; // e.g.: '>= 5.0.0'
```
marekdedic, line 118: I'm a little bit torn about these version constraints - aren't we trying to be too clever here? Maybe this encourages rules that rely on specific things being in specific versions, which IMHO is not a good design... I think I would prefer to have a more limited way of conditioning rules based on Svelte versions - probably just based on major versions, or maybe even version 5 vs. versions 3/4 (just 2 options)...

packages/eslint-plugin-svelte/src/types.ts
```diff
@@ -108,6 +109,18 @@ export interface PartialRuleMetaData {
...
+		fileTypes?: SvelteContext['fileType'][];
+		runes?: SvelteContext['runes'];
+		svelteKitVersion?: string; // e.g.: '>= 1.0.0'
```
marekdedic, line 121: dtto from `svelteVersion`

marekdedic:
Hey @baseballyama, great work on a quick design in one day! I think the general idea is good, I have a lot of questions/notes on the particulars - I left a lot of comments :D

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: Previously [Our site](https://sveltejs.github.io/eslint-plugin-svelte/rules/infinite-reactive-loop/) has browser REPL and it needs to run ESLint in browser.

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: At here, it’s better to keep things simple and store only basic information. This is because the definition of "context" can change depending on the rules. For example, we might want to create a rule that prevents placing files like `xxx.svelte.js` in the same directory as `+page.svelte`.
Another example is that `xxx.svelte.js` might be used for implementing composables, and we might want to enforce a rule that their names must start with `useX`.

In short, since the definition of "context" depends on the rules, it’s better to just manage file types here.

packages/eslint-plugin-svelte/src/utils/index.ts
```diff
@@ -1,4 +1,69 @@
-import type { RuleModule, PartialRuleModule } from '../types.js';
+import type { RuleModule, PartialRuleModule, PartialRuleMetaData, RuleContext } from '../types.js';
+import { getSvelteContext, type SvelteContext } from '../utils/svelte-context.js';
+import semver from 'semver';
+
+function satisfiesCondition(
```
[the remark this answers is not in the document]
baseballyama: Could you explain more?

packages/eslint-plugin-svelte/src/types.ts
baseballyama: This is because making related parts automatically produce type errors when one part is modified makes refactoring easier.

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: updated: https://github.com/sveltejs/eslint-plugin-svelte/pull/980/commits/910effd2057a67921a6ed0b01be5de31aae52d4a

packages/eslint-plugin-svelte/src/types.ts
baseballyama: Agreed and updated: https://github.com/sveltejs/eslint-plugin-svelte/pull/980/commits/910effd2057a67921a6ed0b01be5de31aae52d4a

packages/eslint-plugin-svelte/src/types.ts
baseballyama: updated: https://github.com/sveltejs/eslint-plugin-svelte/pull/980/commits/910effd2057a67921a6ed0b01be5de31aae52d4a

baseballyama:
Now most parts of this should be agreeable. Once this is merged, we can add conditions to all the rules and finally update the recommended config. At that point, the recommended config should be able to include all the recommended rules regardless of the Svelte version or whether SvelteKit is used.

packages/eslint-plugin-svelte/src/types.ts
marekdedic: I don't know, seems to be too clever when it's just a boolean :D But I don't think this is too important :)

packages/eslint-plugin-svelte/src/utils/index.ts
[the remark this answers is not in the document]
marekdedic: I think this function with a long list of ifs being called from a for loop looping over an object is not a good design. I would instead add 5 different functions each checking one condition (instead of the 5 ifs) and then replace the for loop in `shouldRun` with individual calls to them. That'd be much more readable and easy to understand and reason about.

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,201 @@
...
+): Pick<SvelteContext, 'svelteKitFileType' | 'svelteKitVersion'> {
+	const filePath = getFilename(context);
+	const svelteKitVersion = getSvelteKitVersion(filePath);
+	if (svelteKitVersion == null) {
+		return {
+			svelteKitFileType: null,
+			svelteKitVersion: null
+		};
+	}
+	if (isRunInBrowser) {
```
marekdedic, line 87: Sorry to duplicate the comment, the other one was not part of the same review :/

Is the plugin still run in the browser? I thought that functionality was removed. If this is really needed, I think this is not the place to do it - instead the whole `SvelteContext` should be replaced by a mock version when running in the browser playground. That way it is clearly separated from the logic reading things like sveltekit versions...

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,201 @@
+import type { RuleContext } from '../types.js';
+import fs from 'fs';
+import path from 'path';
+import { getPackageJson } from './get-package-json.js';
+import { getFilename, getSourceCode } from './compat.js';
+
+const isRunInBrowser = !fs.readFileSync;
+
+export type SvelteContext = {
+	svelteVersion: '3/4' | 5;
```
marekdedic, line 10: This is a union of a string literal and a number literal - they should both be the same type to make it less confusing for users.

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,201 @@
...
+	fileType: '.svelte' | '.svelte.[js|ts]';
+	runes: boolean;
+	svelteKitVersion: '1-next' | 1 | 2 | null;
```
marekdedic, line 13: dtto about string and number literals.

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: But the design is good!

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: I'm still not convinced. I think we should instead add something like a `fileContext` field where we add some information about the file and we can possibly extend it in the future. I am not sure your first example is possible in eslint, as it operates on files separately. But anyway, we would probably need a separate flag for `.svelte.js/ts` files. But I don't think it is useful to have a way to distinguish `+page.js`, `+page.server.js`, `+layout.js` and `+layout.server.js` - what you would actually want in most situations is 2 flags - "Is this a layout file?" and "Is this a server-only file?"

Does that make sense to you or am I being too convoluted?

marekdedic (changes requested):

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: Hum, Even so, I personally still want to support [vite-plugin-eslint4b](https://github.com/ota-meshi/vite-plugin-eslint4b)

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: I think we can recognize this automatically via TypeScript? I don't have strong opinion for this.

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: This is one of specific example why "Is this a layout file?" and "Is this a server-only file" is not enough.
https://github.com/sveltejs/eslint-plugin-svelte/issues/438

This rule should check only `+layout.(js|ts)`. (`+layout.svelte` should not check.)

packages/eslint-plugin-svelte/src/utils/index.ts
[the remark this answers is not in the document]
baseballyama: updated: https://github.com/sveltejs/eslint-plugin-svelte/pull/980/commits/ad3dfc3ee0a2c4ff80687e7c68c6ecbb037b8312

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: Yes, we can, but not everyone uses TS and I think this is confusing anyway... I'd just make them all strings...

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: Hmm, maybe I'm wrong (I haven't used SvelteKit server-side yet), but #438 should also check `+page.(js|ts)`, no? Then it's exactly what I've been getting at :D The differentiator isn't really a list of filename suffixes, but a role the file plays in SvelteKit...

Anyway, if my change is not accepted, then I would like to think a little bit more about how `fileType` and `svelteKitFileType` interact...

marekdedic: And yes, the list of "file role flags" would probably be longer and would need to expand over time and some rules would check multiple flags at once...

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: Yeah, that makes sense. But then I'd mock the whole `SvelteContext` somewhere else (probably in `getSvelteContext`) and keep this function simple

packages/eslint-plugin-svelte/src/utils/index.ts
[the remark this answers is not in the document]
marekdedic: Thanks, this is better! I have further nitpicks, though (sorry :D):

1. Why is the `isNotSatisfies` function "negated"? Seems like `satisfies` would be simpler to reason about
2. I'm a little bit confused about all the functions here... `isNotSatisfies` checks a single condition, `shouldRun` checks them all, then what does `satisfiesCondition` do?

packages/eslint-plugin-svelte/src/utils/index.ts
```diff
@@ -1,4 +1,50 @@
...
+}
+
+function satisfiesCondition(
+	condition: NonNullable<PartialRuleMetaData['conditions']>[number],
+	svelteContext: SvelteContext
+): boolean {
+	if (
+		isNotSatisfies(svelteContext.svelteVersion, condition.svelteVersions) ||
+		isNotSatisfies(svelteContext.fileType, condition.fileTypes) ||
+		isNotSatisfies(svelteContext.runes, condition.runes) ||
```
marekdedic, line 19: `runes` are a boolean, `isNotSatisfies` doesn't work for that...

marekdedic (changes requested):
Thanks for the changes, unfortunately, I have more questions/comments... But I think we are getting there, good work on this!

packages/eslint-plugin-svelte/src/utils/index.ts
baseballyama: I already changed it to array.

packages/eslint-plugin-svelte/src/utils/index.ts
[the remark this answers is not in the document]
baseballyama: > Why is the isNotSatisfies function "negated"? Seems like satisfies would be simpler to reason about

For me, it’s easier to understand the set if the function name indicates "does not match the condition” rather than something like “no condition or match the condition."

> I'm a little bit confused about all the functions here... isNotSatisfies checks a single condition, shouldRun checks them all, then what does satisfiesCondition do?

I don’t find this particularly hard to understand, but in any case, this is an internal function name, and the type information provides additional context, so I think it’s not a big issue. (I work mainly with typed languages like Java and Kotlin, so I might just be used to it 😅). (Suppose based on your expertise, do you mainly work with Python?)

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: It might just be personal preference, but I see the possibility of this program running in a browser as part of its "spec." That’s why I implemented the `getSvelteKitContext` function to support behavior in the browser. In any case, it’s probably just a matter of preference.

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: I think this is also just personal preference. I find treating numbers as strings can cause confusion, so I’ve always handled numbers as numbers. That said, a union like `"3/4" | 5` can be validated by TypeScript’s type checking, so I don’t see it as a problem. (However, if I'm a reviewer and someone submitted a PR with `"3/4" | "5"`, I wouldn’t object to it.)

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: > And yes, the list of "file role flags" would probably be longer and would need to expand over time and some rules would check multiple flags at once...

Yes. In that case, we can add helper functions for common validation patterns.
I think it’s best to start with basic functionality and then add more convenient features later if they prove to be valuable.

baseballyama: > Hmm, maybe I'm wrong (I haven't used SvelteKit server-side yet), but https://github.com/sveltejs/eslint-plugin-svelte/issues/438 should also check +page.(js|ts),

Yes. We need to check `+page.(js|ts)` also, but what I wanted to say is that "Is this a layout file?" and "Is this a server-only file" is not enough.

baseballyama:
@marekdedic 

Thank you for the review again! Some points that felt like personal preferences were not changed at this time. There are still a few remaining points to discuss, so I’d appreciate your input on those as well. It would be great if we could merge this today. That would bring us significantly closer to v3!

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: I noticed that we’ve repeated the exchange twice. I respect your opinion.

https://github.com/sveltejs/eslint-plugin-svelte/pull/980/commits/531911a246216a0fcbb6993ccb4a12576b01eb7c

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: updated! https://github.com/sveltejs/eslint-plugin-svelte/commit/531911a246216a0fcbb6993ccb4a12576b01eb7c

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: > Yes. In that case, we can add helper functions for common validation patterns.

Hmm, yeah, that's probably a good enough alternative.

I was thinking about the `fileType` a little bit more as well. I get that in `svelteKitFileType`, the list of file suffixes makes sense, but I think it gets confusing with `fileType` pretty quickly - if I understand it correctly, a `+page.svelte` would havea file type `.svelte` and `+page.js` would have a file type of `.svelte.js` - which is a little counterintuitive. Maybe `fileType` could have values `"component"|"js"` or something similar? Then it wouldn't give the illusion of being a filename extension, which it isn't in case of SvelteKit files...

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: > It might just be personal preference, but I see the possibility of this program running in a browser as part of its "spec."

Yeah, I agree. But I don't think this is the right place to implement the check. I think either `shouldRun` should always return true in the browser, or maybe the context could be configurable in the browser environment...

packages/eslint-plugin-svelte/src/types.ts
```diff
@@ -108,6 +109,18 @@ export interface PartialRuleMetaData {
...
 	type: 'problem' | 'suggestion' | 'layout';
+	/**
+	 * Conditions to determine whether this rule should be applied.
+	 * Multiple conditions can be specified as array, and the rule will be applied if any one of them matches (logical OR).
+	 * If not specified, the rule will be applied to all files.
+	 */
+	conditions?: {
+		svelteVersions?: SvelteContext['svelteVersion'][];
+		fileTypes?: SvelteContext['fileType'][];
+		runes?: SvelteContext['runes'][];
```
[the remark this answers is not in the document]
marekdedic: (I get that the check is then easier to implement, but that shouldn't be the deciding factor)

packages/eslint-plugin-svelte/src/utils/index.ts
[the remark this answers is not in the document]
marekdedic: Sorry, the second part of my last comment was wrong - I didn't realize `conditions` was an array of objects, not just one object...

marekdedic:
Hi @baseballyama, I think we are getting there (honestly now it's just a few small details). Thanks for taking the time!

packages/eslint-plugin-svelte/src/types.ts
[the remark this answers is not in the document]
baseballyama: What if `null` is introduced for `runes` in the future? `runes` can also be an array, so that if `runes` is no longer binary in the future, we can follow suit without breaking changes. For now, there is no necessity for it to be an array, but I don't think there is any necessity for it not to be an array either.

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: I don't think this is the scope of this PR. Also, since the code for browsers is not much complex and not a major use case at this time, it should be modified when this need is further increased or the complexity rises. If you have suggestions for improvement, we would welcome another PR.

packages/eslint-plugin-svelte/src/utils/index.ts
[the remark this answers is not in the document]
baseballyama: I wanted to make it a little shorter, so after discuss with ChatGPT, I decided to use `doesNotSatisfy`.

https://github.com/sveltejs/eslint-plugin-svelte/pull/980/commits/a36d9359a39e91fcea931a34be5ee377ad5a2a08

packages/eslint-plugin-svelte/src/types.ts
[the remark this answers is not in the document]
marekdedic: Sorry, I don't understand what `null` being introduces for runes means. Runes are currently either on or off. While it is theoretically possible that in the future there may be some "Runes v2", I think that's a pretty big stretch and it's much more likely that everything will convert to runes in the future. If I understand it correctly, `[true, false]` is functionally equal to `undefined`, so the only 2 practically useful values (apart from `undefined`) are `[true]` and `[false]`. That is confusing for the users. And if you want to be specific, you can explicitly pass `undefined`...

marekdedic (changes requested):
One final thing and then we're done :)

packages/eslint-plugin-svelte/src/types.ts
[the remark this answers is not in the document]
baseballyama: I updated it because what I’m saying can actually happen if a user uses different parser.

https://github.com/sveltejs/eslint-plugin-svelte/pull/980/commits/afd422d9e78086bb2a2ca3dffc65a23d6e4e4c48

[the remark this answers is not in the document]
baseballyama: I made it more robust:

https://github.com/sveltejs/eslint-plugin-svelte/pull/980/commits/f686b4da9bf5f9620cd3971917a365e87302b768

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,259 @@
...
+import path from 'path';
+import { getPackageJsons } from './get-package-json.js';
+import { getFilename, getSourceCode } from './compat.js';
+
+const isRunInBrowser = !fs.readFileSync;
+
+export type SvelteContext = (
+	| ({
+			svelteVersion: '3/4';
+	  } & {
```
marekdedic, line 12: Why is this split into 2?

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,259 @@
...
+			svelteFileType: '.svelte' | null;
+			runes: null;
```
marekdedic, line 14: Why `null` here and not `false`? This conflates it with a non-svelte file, rather than a Svelte 5 file in non-runes mode, which I think is semantically much closer..

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,259 @@
...
+	  })
+	| ({
+			svelteVersion: '5';
+	  } & (
+			| {
+					svelteFileType: '.svelte' | '.svelte.[js|ts]';
```
marekdedic, line 20:
```suggestion
					svelteFileType: '.svelte' | '.svelte.js';
```
In `svelteKitFileType`, TS extensions are omitted, I'd omit them here as well... (or add them there, either way, I'd do the same thing for both fields)

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
```diff
@@ -0,0 +1,259 @@
...
+					/** If a user uses a parser other than `svelte-eslint-parser`, `undetermined` will be set. */
+					runes: boolean | 'undetermined';
```
marekdedic, line 22: Hmm, I'm having a hard time understanding the difference between `null` and `'undetermined'` here - they both mean we couldn't figure it out, no?

marekdedic (commented):
The new `SvelteContext` is much clearer, thanks! Unfortunately, that means there's new things for me not to understand :D

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: For more strict typing. Reinforcing the types eliminates improbable patterns.

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: Because Svelte 3/4 doesn't have runes. `false` means "There is a concept of runes, but it is not applied in this file" but if that concept does not exist, I uses `null`.
Anyway we can recognize this via typing.

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: changed it to js|ts.
https://github.com/sveltejs/eslint-plugin-svelte/pull/980/commits/4640ef535d92fdee4a404d1a74acedfaca22fcad

It states that it does not support coffeescript for now. (If someone raises an issue in the future, we may respond to it.)

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: We can automatically recognize this via typing.

- null: The concept of runes itself does not exist.
- true: Using runes
- false: Not using runes
- undetermined: The concept of runes exists but I don't know if it is used

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: Ok, this works :)

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: I don't understant, I thought `{a: string} & {b: int}` is equal to {a: string; b: int}`, isn't it?

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
marekdedic: Ah, ok. Is there any menaingful difference between `runes don't exist` and `not using runes`? Because they semm to me to be equivalent in all useful aspects...

baseballyama: I think only facts should be expressed here, as usefulness depends on context

packages/eslint-plugin-svelte/src/utils/svelte-context.ts
baseballyama: updated: https://github.com/sveltejs/eslint-plugin-svelte/pull/980/commits/0e28cc19edc9b33c6d96abe64bced627f7851e1f

baseballyama:
@marekdedic 

Thank you for taking the time to review this! I think we are in agreement on almost areas and will merge to move forward.👍

marekdedic:
Great, thanks for the work on this, I think this design is pretty great now :)

PR #1005, opened 2025-01-11
chore: execute svelte/experimental-require-slot-types only on Svelte 3 and 4

baseballyama:


baseballyama:
I was doing the wrong thing.

PR #1006, opened 2025-01-12
chore: do not run `infinite-reactive-loop` rule on Svelte 5 with runes

baseballyama:
Related https://github.com/sveltejs/eslint-plugin-svelte/issues/894.
But add a configuration to disable rules that are unnecessary when running with Svelte 5 Runes.


ota-meshi (approved):
LGTM! Thank you!

PR #1007, opened 2025-01-12
fix: update method for extracting major version

baseballyama:
This PR addresses a bug in the method for extracting the library’s major version. While it does not support all possible version specification methods, I believe it is sufficient for extracting the major version of Svelte and SvelteKit.
This bug is introduced in https://github.com/sveltejs/eslint-plugin-svelte/pull/980.


baseballyama:
Merge this because I need this for https://github.com/sveltejs/eslint-plugin-svelte/pull/1006

PR #1012, opened 2025-01-12
chore: change the rule category of `no-export-load-in-svelte-module-in-kit-pages` to `SvelteKit`

baseballyama:


ota-meshi (approved):
LGTM! Thank you!

PR #1013, opened 2025-01-12
feat: add Svelte 5 support to `no-not-function-handler`

baseballyama:
I updated docs also. (`on:click` -> `onclick`)

image


packages/eslint-plugin-svelte/src/rules/no-not-function-handler.ts
```diff
@@ -95,6 +95,18 @@ export default createRule('no-not-function-handler', {
 					return;
 				}
 				verify(node.expression);
+			},
+			SvelteAttribute(node) {
+				if (node.key.type === 'SvelteName' && node.key.name.startsWith('on')) {
```
ota-meshi, line 100: I don't know if that's actually the case, but I'm worried that there might be false positives for components with properties like `one`, `only`, `once`, and `online`. 🤔 

I think maybe we should ignore component attributes from the check. What do you think?

baseballyama: Yes. We need to tale every events from here.
https://github.com/sveltejs/svelte/blob/a129592e5b248a734a68da6e9028941803a3d063/packages/svelte/elements.d.ts#L83

baseballyama: Did! https://github.com/sveltejs/eslint-plugin-svelte/pull/1013/commits/2e2b6f0fadf6b539a3da29a0d4634cd5633c711e

ota-meshi (approved):
LGTM! Thank you!

PR #1017, opened 2025-01-12
chore: do not run `no-reactive-functions` rule on Svelte 5 with runes

baseballyama:


baseballyama:
Should we write code that supports Svelte 5?

ota-meshi:
I think this PR change is enough for the rule 😃

ota-meshi (approved):
LGTM!
No more `WIP`, right?

PR #1018, opened 2025-01-12
chore: run the `no-inspect` rule only in Svelte 5’s runes mode

baseballyama:


baseballyama:
https://github.com/sveltejs/svelte-eslint-parser/pull/636 fixes CI.

Issue #785, opened 2024-06-14
Addition of a config that includes best practice rules.


tkhs0813:
### Description

The currently provided config does not include all [best practices](https://sveltejs.github.io/eslint-plugin-svelte/rules/#best-practices).
I want a stronger config than plugin:svelte/recommended. Is it possible to add this?

Thank you for considering my request. I believe this feature would be a valuable addition to the project and help improve the overall user experience.

Best regards,

ota-meshi:
Thank you for posting this issue.

The recommended configuration provided by the plugin is determined by our preferences as maintainers. But, the rules enabled in that configuration are probably those that are essential for most people.

Maybe your needs will require a configuration that provides a new set of rules for users who want stricter checking.
However, in my opinion, coming up with a new set of rules would be time-consuming and would put a lot of strain on us as maintainers. To avoid it, I think it would be better if someone provided a new configuration as a separate project.

What do you think?

baseballyama:
I think the reason rules weren’t added to the recommended config is that we followed ESLint’s versioning policy. Since we changed the versioning policy starting from v3, this shouldn’t be a big issue anymore.

Issue #755, opened 2024-05-09
Clean up configs


baseballyama:
Currently we have 4 configs.

- plugin:svelte/base
- plugin:svelte/recommended
- plugin:svelte/prettier
- plugin:svelte/all

But some rules is may needed for only Svelte 4, Svelte 5 without runes, Svelte 5 with runes.

Therefore, I think it is better to release a major version of `eslint-plugin-svelte` and clean up configs when Svelte 5 is GA.

I will sort out the specific config later, but first I would like to confirm that this policy seems to be a good or not.

ota-meshi:
Agree 👍  I think it would be good if we could provide configurations that match the user's usage.

PR #1033, opened 2025-01-19
chore!: update recommended config

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/755
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/785
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/894


baseballyama:
@ota-meshi 
The current configuration seems quite aggressive. Could you share your thoughts on which rules should be included as recommended?

marekdedic:
Hi, thanks, I am thinking about whether these shouldn't be added as well (in decreasing order of certainty):
- `svelte/no-navigation-without-base` - This one warns about what is incorrect code (though it may be fixed in SvelteKit at some point in the future)
- `svelte/button-has-type` - This one is about what is considered best practice
- `svelte/no-unused-class-name` - This one is more about unused code...

baseballyama:
> svelte/no-navigation-without-base

I don’t think this needs to be added. (In fact, it’s better not to add it.) This is because it would result in meaningless linting errors for users who don’t use basepath. Only users who use basepath should opt in to this rule.

> svelte/button-has-type

This is not included in the recommended rules for Vue either. IIRC it was added to the Vue's recommended config once before but was removed due to user pushback.
https://github.com/vuejs/eslint-plugin-vue/pull/1432

> svelte/no-unused-class-name

This was initially added, but in the codebases I tested, false positives occurred. Therefore, I judged that it was not mature enough to be included in the recommended rules.

baseballyama:
> I am torn on this one - on the one hand, it would be more warnings for users, on the other, I consider not adding the base path to be incorrect - if the code is e.g. reused in the future, this is a serious foot-gun (So much so that the SvelteKit team considers fixing it on their end).

IMO:
Whether to reuse the code with a different basepath is up to each project. If there’s no chance of reuse, we don’t need to consider the basepath at all. If it turns out later that a basepath is needed, we can simply opt into this rule and make adjustments then.

Code is read more often than it is written, so reducing unnecessary noise is important. For projects where the basepath doesn’t need to be considered, the code added by this rule would just become noise.

ota-meshi:
The config provided by eslint-plugin-vue follows the Vue style guide, so if something is not mentioned in the style guide, it will not be added to the config. Therefore, the `vue/button-has-type` rule is not included in the config.

However, neither `react/button-has-type` nor `@angular-eslint/template/button-has-type` seem to be included in the config provided by those plugins.
Therefore, it seems reasonable to not include `svelte/button-has-type` in the config either.

baseballyama:
I believe we’ve agreed on the recommended config. Finally, I’ll document the changes in detail in the changeset and wrap it up.

ota-meshi (approved):
Thank you very much!

PR #1034, opened 2025-01-19
chore: add Svelte 5 docs for `prefer-destructured-store-props` rule

baseballyama:


marekdedic:
Why this? You can still use stores with runes?

baseballyama:
Does the performance differ depending on whether the GOOD or BAD syntax is used when Svelte 5 is run in runes mode?

```
<script>
  /* eslint svelte/prefer-destructured-store-props: "error" */
  import store from './store.js';
  $: ({ foo } = $store);
</script>

<!-- ✓ GOOD -->
{foo}

<!-- ✗ BAD -->
{$store.foo}
```

baseballyama:
Further investigation confirmed that using code like the following for stores in Svelte 5 does provide better performance:

```ts
let foo = $derived($store.foo)
```

baseballyama:
So I updated the docs instead of changing the config.

ota-meshi (approved):
LGTM! Thank you!

PR #1038, opened 2025-01-22
feat: set `languageOptions.globals` in base config

TatsuyaYamamoto:
`eslint-plugin-svelte`'s config does not have `languageOptions.globals`.
https://github.com/sveltejs/eslint-plugin-svelte/blob/main/packages/eslint-plugin-svelte/src/configs/flat/base.ts

I think many Svelte applications require `globals.browser`, so is it worth providing it in the base config?

references:
- `docs-svelte-kit` set `globals.browser`: https://github.com/sveltejs/eslint-plugin-svelte/blob/main/docs-svelte-kit/eslint.config.mjs#L34
- `eslint-plugin-vue` provides `globals.browser` in base config: https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/configs/flat/base.js#L17

### DIFF
--- a/packages/eslint-plugin-svelte/package.json
+++ b/packages/eslint-plugin-svelte/package.json
@@ -60,6 +60,7 @@
     "@jridgewell/sourcemap-codec": "^1.5.0",
     "eslint-compat-utils": "^0.6.4",
     "esutils": "^2.0.3",
+    "globals": "^15.14.0",
     "known-css-properties": "^0.35.0",
     "postcss": "^8.4.49",
     "postcss-load-config": "^3.1.4",
--- a/packages/eslint-plugin-svelte/src/configs/flat/base.ts
+++ b/packages/eslint-plugin-svelte/src/configs/flat/base.ts
@@ -2,6 +2,7 @@
 // This file has been automatically generated,
 // in order to update its content execute "pnpm run update"
 import type { ESLint, Linter } from 'eslint';
+import globals from 'globals';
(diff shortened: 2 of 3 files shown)


baseballyama:
Is my understanding correct that setting `globals` is necessary for frontend projects regardless of whether `eslint-plugin-svelte` is used or not? Even if `eslint-plugin-svelte` includes `globals`, other plugins or core rules that don’t rely on this plugin would still require `globals`.

In that case, users would need to configure `globals` anyway, meaning this PR doesn’t change what they need to do. Is that correct? @ota-meshi

ota-meshi:
I don't know if it's necessary for all users to use `globals.browser`.


I think `globals.browser` is necessary because `*.svelte` files run on the browser, but users who use SSR may not want to set `globals.browser`. I think users might prefer `/* globals document */` depending on their project.
I think it might be better to mention `globals.browser` in the documentation instead of providing it from our config, because it's easier to turn on multiple identifiers from globals than to turn them off.
What do you think?

baseballyama:
This is exactly what I want to say.
Already I added this in https://github.com/sveltejs/eslint-plugin-svelte/pull/1041/

https://github.com/sveltejs/eslint-plugin-svelte/pull/1041/files#diff-971a69e5ae264bfdae535d0c72902cc6f15a559f6d2dc650b78b473f6783c3caR45-R46

So close this PR now.
Thank you for your contribution @TatsuyaYamamoto !

Issue #1036, opened 2025-01-20
Don't emit "svelte/no-useless-mustaches" on multiline template literals


lishaduck:
### Description

I have a component that wraps `textarea` with Markdown support. Because I can't pass strings to children (is that true? I'm new to Svelte), I had to make a `value` prop.
When I pass a markdown string in, I get a `svelte/no-useless-mustaches` warning, but the mustache isn't useless as the template literal allows me to make it multiline.

```
<HyperMD value={`# Doc

Text goes here`} />
```

I know I could just move it, but I feel like the rule is wrong here, the mustache isn't useless.

PR #1040, opened 2025-01-26
fix: account for \n and \r in `TemplateLiteral` handling for the `no-useless-mustaches` rule

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1036


ota-meshi (approved):
LGTM! Thank you!

PR #1043, opened 2025-01-26
feat: add `*.svelte.{js,ts}` config to base config

### LINKED ISSUE #1030: No runes outside svelte
### Motivation

I don't think an explanation is needed - using runes outside Svelte files results in runtime errors/crashes, or at the very least doesn't work.

### Description

No runes at all outside `.svelte`, `.svelte.{js,ts}` files (per configuration), except if imported, i.e. disallow global use of rune identifiers in regular files.

ota-meshi:


baseballyama:
I think this PR is helpful, but I don’t this it can close #1030.
To put it simply, #1030 is asking for a rule to report cases where runes like `$state` are used in `.js` files (not `.svelte.js` files).

ota-meshi:
Thanks for letting me know! I guess I misunderstood 😅
I removed the `close` comment.

baseballyama (approved):

Issue #1045, opened 2025-01-27
Support Svelte 5 in `valid-prop-names-in-kit-pages`


lishaduck:
### Description

Layouts in Svelte 5 use a `children` prop, which `valid-prop-names-in-kit-pages` (newly added to the recommended set in 3.0.0-next.16) flags

(No, this isn't really a feature request, but it's not quite a bug either.)

baseballyama:
I think I already support Svelte5. Could you provide REPL?

PR #1048, opened 2025-02-01
fix: allow `children` in `valid-prop-names-in-kit-pages` rule

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1045


ota-meshi (approved):
LGTM! Thank you!

PR #1054, opened 2025-02-05
feat: implemented the `valid-style-parse` rule

### LINKED ISSUE #509: Add rule checking for successful style parsing
### Motivation

With the recent edition of a rule that uses the parsed styles (#489), this rule in fact doesn't produce any errors if the styles cannot be parsed (either they're invalid, or `svelte-eslint-parser` cannot parse the used style language)

### Description

I propose a rule that would produce a warning in either of these 2 cases (parse error / unknown language). This would guarantee that rules using styles (currently only `no-unused-class-name`) work correctly. There's also a minor overlap with the `block-lang` rule, but I think this is OK

### Examples

```svelte
<!-- ✓ GOOD -->
<style>
  .class {
    font-weight: bold;
  }
</style>

<style lang="scss">
  .class {
    font-weight: bold;
  }
</style>

<!-- ✗ BAD -->
<style>
  .class
    font-weight: bold;
</style>

<style lang="unknown">
  .class {
    font-weight: bold;
  }
</style>
```

### Additional comments

_No response_

marekdedic:
Closes #509

### DIFF
--- a/docs/rules/valid-style-parse.md
+++ b/docs/rules/valid-style-parse.md
@@ -1,0 +1,78 @@
+---
+pageClass: 'rule-details'
+sidebarDepth: 0
+title: 'svelte/valid-style-parse'
+description: 'require valid style element parsing'
+---
+
+# svelte/valid-style-parse
+
+> require valid style element parsing
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> **_This rule has not been released yet._** </badge>
+
+## :book: Rule Details
+
+This rule reports issues with parsing of the `<style>` element by the svelte-eslint-parser.
+
+<!--eslint-skip-->
+
+```svelte
+<script>
+  /* eslint svelte/valid-style-parse: ["error"] */
+</script>
+
+<!-- ✓ GOOD -->
+<style>
+  .class {
+    font-weight: bold;
+  }
+</style>
+```
+
+```svelte
+<script>
+  /* eslint svelte/valid-style-parse: ["error"] */
+</script>
+
+<!-- ✓ GOOD -->
+<style lang="scss">
+  .class {
+    font-weight: bold;
+  }
+</style>
+```
+
+```svelte
+<script>
+  /* eslint svelte/valid-style-parse: ["error"] */
+</script>
+
+<!-- ✗ BAD -->
+<style>
+  .class
+    font-weight: bold;
+</style>
+```
+
+```svelte
+<script>
+  /* eslint svelte/valid-style-parse: ["error"] */
+</script>
+
+<!-- ✗ BAD -->
+<style lang="unknown">
+  .class {
+    font-weight: bold;
+  }
+</style>
+```
+
+## :wrench: Options
+
+Nothing.
+
+## :mag: Implementation
+
+- [Rule source](https://github.com/sveltejs/eslint-plugin-svelte/blob/main/packages/eslint-plugin-svelte/src/rules/valid-style-parse.ts)
+- [Test source](https://github.com/sveltejs/eslint-plugin-svelte/blob/main/packages/eslint-plugin-svelte/tests/src/rules/valid-style-parse.ts)
--- a/packages/eslint-plugin-svelte/src/rules/valid-style-parse.ts
+++ b/packages/eslint-plugin-svelte/src/rules/valid-style-parse.ts
@@ -1,0 +1,40 @@
+import { createRule } from '../utils/index.js';
+import { getCwd, getSourceCode } from '../utils/compat.js';
+
+export default createRule('valid-style-parse', {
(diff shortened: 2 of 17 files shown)


marekdedic:
There is some issue in the pkg.pr.new action, I don't think it's related to anything in this PR...

ota-meshi:
The workflow has been fixed 😄 
I've rebased this PR so please apply it to your local too.

docs/rules/valid-style-parse.md
```diff
@@ -0,0 +1,58 @@
...
+</style>
+```
+
+## :wrench: Options
+
+```json
+{
+  "svelte/valid-style-parse": ["error", {}]
+}
+```
```
ota-meshi, line 53:
```suggestion
Nothing.
```

docs/rules/valid-style-parse.md
```diff
@@ -0,0 +1,58 @@
...
+
+# svelte/valid-style-parse
+
+> require valid style element parsing
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> **_This rule has not been released yet._** </badge>
+
+## :book: Rule Details
+
+This rule reports issues with parsing of the `<style>` element by the ESLint Svelte parser.
```
ota-meshi, line 16:
```suggestion
This rule reports issues with parsing of the `<style>` element by the svelte-eslint-parser.
```

packages/eslint-plugin-svelte/src/rules/valid-style-parse.ts
```diff
@@ -0,0 +1,41 @@
...
+			return {};
+		}
+
+		return {
+			SvelteStyleElement(node) {
+				const styleContext = sourceCode.parserServices.getStyleContext!();
+				if (styleContext.status === 'parse-error') {
+					context.report({
+						loc: node.loc,
+						messageId: 'parseError'
```
ota-meshi, line 29: Can we show what the parsing error is?

ota-meshi (changes requested):
Thank you for this PR!
It looks mostly good to me, but I have a few comments.

docs/rules/valid-style-parse.md
```diff
@@ -0,0 +1,58 @@
...
+  .class
+    font-weight: bold;
+</style>
+
+<style lang="unknown">
+  .class {
+    font-weight: bold;
+  }
+</style>
+```
```
ota-meshi, line 45: Could you write the following so that Twoslash ESLint works?

````suggestion
```svelte
<script>
  /* eslint svelte/valid-style-parse: ["error"] */
</script>
<!-- ✓ GOOD -->
<style>
  .class {
    font-weight: bold;
  }
</style>
```

```svelte
<script>
  /* eslint svelte/valid-style-parse: ["error"] */
</script>
<!-- ✓ GOOD -->
<style lang="scss">
  .class {
    font-weight: bold;
  }
</style>
```

```svelte
<script>
  /* eslint svelte/valid-style-parse: ["error"] */
</script>
<!-- ✗ BAD -->
<style>
  .class
    font-weight: bold;
</style>
```

```svelte
<script>
  /* eslint svelte/valid-style-parse: ["error"] */
</script>
<!-- ✗ BAD -->
<style lang="unknown">
  .class {
    font-weight: bold;
  }
</style>
```
````

File: packages/eslint-plugin-svelte/src/rules/valid-style-parse.ts (hunk 3 above)
Comment (marekdedic) [reply packages/eslint-plugin-svelte/src/rules/valid-style-parse.ts]: Great idea, done :)

**Comment by marekdedic** [comment]
@ota-meshi I can't solve the update error - The update script fails on the invalid CSS, which makes sense. I tried wrapping the `writeAndFormat` function with a try/catch block, but for some reason, the error isn't getting caught there. I suspect it is actually caught somewhere in prettier and printed to stderr directly from there..

BTW, that function can be simplified to:
```ts
export async function writeAndFormat(fileName: string, content: string): Promise<void> {
	fs.writeFileSync(fileName, content, 'utf8');
	const prettierrc = await prettier.resolveConfig(fileName);
	const formatted = await prettier.format(content, { filepath: fileName, ...prettierrc });
	fs.writeFileSync(fileName, formatted, 'utf8');

	const eslint = new ESLint({ fix: true });
	const [result] = await eslint.lintText(formatted, { filePath: fileName });
	if (result.output) {
		fs.writeFileSync(fileName, result.output, 'utf8');
	}
}
```

ota-meshi:
It seems that prettier will output a log and continue processing even if it fails to parse the style.
If we want to avoid logging, we will need to write `<!-- prettier-ignore -->` above the code block. However, prettier works, so it is not necessary for now.

ota-meshi (approved):
LGTM! Thank you!

PR #1055, opened 2025-02-05
chore: update pkg.pr.new workflow

ota-meshi:
This PR changes the workflow of pkg.pr.new.

The workflow of #1054 is failing with 403 error.
I expect that by splitting the workflow in this PR based on the svelte repository, we can avoid the 403 error.


ota-meshi:
Hmm... It seems the new workflow isn't being called.

ota-meshi:
It seems that the workflow by workflow_run will not work unless we merge this.
I'll try merging this.

Issue #1052, opened 2025-02-03
disallow $state(new SvelteSet) and similar


mustafa0x:
### Motivation

SvelteMap, SvelteSet, and SvelteURL don't need to be wrapped in $state. but that's not always obvious

### Description

per above

### Examples

```svelte
<!-- ✓ GOOD -->
<script>
const myset = new SvelteSet
</script>

<!-- ✗ BAD -->
<script>
const myset = $state(new SvelteSet)
</script>
```

marekdedic:
Hmm, reacting quickly and not thoroughly enough, but this needs a little bit more nuance:

```ts

let myOptionalSet = $state(new SvelteSet());

// Somewhere further down
myOptionalSet = null; // This needs the variable to be wrapped in a $state in order for this to be reactive.
```

mikededo:
I agree with @marekdedic. The statement that reactive built-ins _do not need to be wrapped_ with `$state` is not correct.

marekdedic:
Yes, at the same time, it actually is a mistake like _95% of the time_... So I think it's a great candidate for a lint rule, the rule just needs to be clever and actually figure out when the `$state` is really not needed.

Issue #910, opened 2024-11-08
Add rule `require-event-prefix`


marekdedic:
### Motivation

In Svelte 5, all HTML element events are props starting with "on". Component events can have any name, so it is not clear, which props are events and which are not.

### Description

I'd like to add a rule that requires that all props that are events (i.e. that are functions) start with "on"

### Examples

```svelte
<script lang="ts">
  /* ✓ GOOD */

  interface Props {
    regularProp: string;
    onclick(): void;
  }

  let { regularProp, onclick }: Props = $props;

  /* ✗ BAD */

  interface Props {
    click(): void;
  }

  let { click }: Props = $props;
</script>
```

### Additional comments

I think this works even without TS, as the type would be in the parsed AST, right?

ota-meshi:
Hmm. I think it would be useful to have that rule, but I'm worried about whether it's okay to judge all function props as event props 🤔.

What about implementing a rule that checks naming conventions for all props and provides examples for event names in the rule's documentation?
The rule could also disallow the `on` prefix for non-function props.

marekdedic:
Well, in Svelte 5, IIUC there are no special event props - it's all just props, functions or not... I think in that case, it makes sense to treat function props as events...

Disallowing the on prefix for non-function props makes sense to me :+1: 

I don't really understand your point about naming convention for all props?

ota-meshi:
>  I don't really understand your point about naming convention for all props?

For example, the following rule:

```jsonc
{
  "svelte/props-naming-convention": [
    "error",
    {
      "pattern": [
        "/^(?!on)/u" // Non-function props should not start with `on`.
      ],
      "overrides": [
        {
          "type": "function",
          "pattern": [
            "/^on/u" // Function props must start with `on`.
          ]
        },
      ]
    }
  ]
}
```

```svelte
<script>
  /* ✓ GOOD */
  let { onclick = () => {}, num = 42 } = $props();
  /* ✗ BAD */
  let { onFoo = 42, bar = () => {} } = $props();
</script>
```


```jsonc
{
  "svelte/props-naming-convention": [
    "error",
    {
      "pattern": [...],
      "overrides": [
        {
          "type": "boolean",
          "pattern": [
            "/^is/u" // Boolean props must start with `is`.
          ],
        },
      ]
    }
  ]
}
```

```svelte
<script>
  /* ✓ GOOD */
  let { isEnable = true } = $props();
  /* ✗ BAD */
[...]

PR #1062, opened 2025-02-09
feat: add `no-unnecessary-state-wrap` rule

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1052


marekdedic:
Hi,
sorry if this is unfinished and you're planning to add this, but atm, this ignores https://github.com/sveltejs/eslint-plugin-svelte/issues/1052#issuecomment-2630480779

baseballyama:
Since the `clear` function can be used, `$state` is usually unnecessary. If there is a reason to use `$state(new SvelteMap())`, this rule can simply be disabled. ESLint is fully configurable.

marekdedic:
I still have to disagree - the rule now has a clear false positive that can be detected by its implementation, but just isn't.
And I'm sorry, but yeah, you can disable any rule for any line, but if the FP can be detected by the rule, it should be. Only when it's not possible (like what we're discussing in #910 ) should users be forced to manually disable rules...

baseballyama:
Can you provide an example of fp?

marekdedic:
Of course, see https://github.com/sveltejs/eslint-plugin-svelte/issues/1052#issuecomment-2630480779 for a simple example or [here](https://github.com/scout-handbook/handbook-web-admin/blob/2bd402626049e75af7a6d7b9c0b1f7a75b6bff2e/src/lib/resources/fields.svelte.ts#L11-L25) for a real-world example (this one is in `$derived`, but the same semantics apply):

baseballyama:
Then, let’s add an option to not report an error when a reassignment exists.

marekdedic:
Thanks :) I don't even think it needs to be an option - if a `$state(SvelteMap)` is reassigned, the `$state` is always necessary...

baseballyama:
I don’t think so. I believe it might simply be a mistake. At least in my project, I want to consistently treat it as an error. For example, in the following case, it can clearly use “clear”.

```js
  let map = $state(new SvelteMap());
  map = new SvelteMap();
```

marekdedic:
Hmm, you are absolutely right that the code snippet is incorrect, thanks, I din't realize that!

I am thinking if this isn't a separate issue, though - the anti-pattern here is IMO assigning a `new SvelteMap()` to an object which is a `SvelteMap` instead of using `.clear()`. And that would also be an issue with the regular `Map`, wouldn't it? So maybe you found a more general issue here that assigning `new X()` to a variable of type `X` is incorrect and can be replaced with `X.clear()` for X being any of `SvelteMap`, `SvelteSet`, `Map`, `Set` and possibly more...

Now if that were a separate rule it would nicely dovetail with this one even if this rule ignored all re-assigned variables, the other rule would replace the re-assignment with `.clear()` and this rule would then remove the `$state`.

I feel I am rambling a little bit - do I make any sense to you? :D

baseballyama:
In any case, for those who don’t want this rule enabled, they can simply disable it. In most cases, `$state(new SvelteMap())` doesn’t really make sense, so I think having it enabled by default isn’t a bad idea.

baseballyama:
I added `allowReassign` option.

ota-meshi (approved):
LGTM! Thanks!

PR #1063, opened 2025-02-09
feat: detect version information from `node_modules` for pnpm’s catalog

baseballyama:
I noticed this issue while using pnpm catalog in our project.

### DIFF
--- a/.changeset/two-poets-wait.md
+++ b/.changeset/two-poets-wait.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': minor
+---
+
+feat: detect version information from node_modules when not specified in package.json, like pnpm’s catalog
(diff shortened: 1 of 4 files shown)


baseballyama:
I have confirmed that it works correctly in my project using pnpm catalog.

ota-meshi:
I just realized, how about getting the version like this:

```js
import  {VERSION} from "svelte/compiler"
```

I think it will also work with Svelte v3.

baseballyama:
> import  {VERSION} from "svelte/compiler"

I can’t believe it. How did I overlook this until now 😇. I have confirmed that this works even with the 5.0.0-next version.
https://www.npmjs.com/package/svelte/v/5.0.0-next.272?activeTab=code

baseballyama:
I updated the PR according to your suggegstion.

PR #1064, opened 2025-02-09
feat: add `excludedRunes` option to the `prefer-const` rule

baseballyama:
close part of https://github.com/sveltejs/eslint-plugin-svelte/issues/818

As explained here, `$state` also needs to be excluded from the `prefer-const` rule.
https://github.com/sveltejs/eslint-plugin-svelte/pull/985#issuecomment-2566907151

### DIFF
--- a/docs/rules/prefer-const.md
+++ b/docs/rules/prefer-const.md
@@ -14,7 +14,7 @@ since: 'v3.0.0-next.6'
 
 ## :book: Rule Details
 
-This rule reports the same as the base ESLint `prefer-const` rule, except that ignores Svelte reactive values such as `$derived`, `$state`, and `$props`. If this rule is active, make sure to disable the base `prefer-const` rule, as it will conflict with this rule.
+This rule reports the same as the base ESLint `prefer-const` rule, except that ignores Svelte reactive values such as `$derived` and `$props` as default. If this rule is active, make sure to disable the base `prefer-const` rule, as it will conflict with this rule.
 
 <!--eslint-skip-->
 
--- a/docs/rules/prefer-const.md
+++ b/docs/rules/prefer-const.md
@@ -14,7 +14,7 @@ since: 'v3.0.0-next.6'
 
 ## :book: Rule Details
 
-This rule reports the same as the base ESLint `prefer-const` rule, except that ignores Svelte reactive values such as `$derived` and `$props`. If this rule is active, make sure to disable the base `prefer-const` rule, as it will conflict with this rule.
+This rule reports the same as the base ESLint `prefer-const` rule, except that ignores Svelte reactive values such as `$derived`, `$state`, and `$props`. If this rule is active, make sure to disable the base `prefer-const` rule, as it will conflict with this rule.
 
 <!--eslint-skip-->
 
--- a/docs/rules/prefer-const.md
+++ b/docs/rules/prefer-const.md
@@ -46,7 +46,8 @@ This rule reports the same as the base ESLint `prefer-const` rule, except that i
     "error",
     {
       "destructuring": "any",
-      "ignoreReadonly": true
+      "ignoreReadonly": true,
+      "excludedRunes": ["$props", "$state"]
     }
   ]
 }
@@ -56,6 +57,7 @@ This rule reports the same as the base ESLint `prefer-const` rule, except that i
   - `any` (default): if any variables in destructuring should be const, this rule warns for those variables.
   - `all`: if all variables in destructuring should be const, this rule warns the variables. Otherwise, ignores them.
 - `ignoreReadonly`: If `true`, this rule will ignore variables that are read between the declaration and the _first_ assignment.
+- `excludedRunes`: An array of rune names that should be ignored. Even if a rune is declared with `let`, it will still be ignored.
 
 ## :books: Further Reading
 
--- a/docs/rules/prefer-const.md
+++ b/docs/rules/prefer-const.md
@@ -47,7 +47,7 @@ This rule reports the same as the base ESLint `prefer-const` rule, except that i
     {
       "destructuring": "any",
       "ignoreReadonly": true,
-      "excludedRunes": ["$props", "$state"]
+      "excludedRunes": ["$props", "$derived"]
     }
   ]
 }
--- a/.changeset/sixty-cars-fail.md
+++ b/.changeset/sixty-cars-fail.md
@@ -1,5 +1,5 @@
 ---
-'eslint-plugin-svelte': patch
+'eslint-plugin-svelte': minor
 ---
 
-fix: ignore `$state` in the `prefer-const` rule
+feat: add `excludedRunes` option to the `prefer-const` rule
--- a/.changeset/sixty-cars-fail.md
+++ b/.changeset/sixty-cars-fail.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': patch
+---
+
+fix: ignore `$state` in the `prefer-const` rule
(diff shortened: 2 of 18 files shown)


ota-meshi:
I still don't understand why you want to ignore `$state` from `svelte/prefer-const` rule.

I know that some users want to use `let` to change `value` with `const value = $state({ value: new Date().toString() });`, but I think that user would do the same with `const value = { value: new Date().toString() };` which doesn't use `$state()`. In other words, I think that user would always turn off the `prefer-const` rule.
So I don't think there's a need to ignore `$state` from the `svelte/prefer-const` rule check.

baseballyama:
We might want to enable `prefer-const` for plain JavaScript that has nothing to do with Svelte.

In other words, semantically, since `$state` is something that can be reassigned, we may want to consistently use `let` for it. However, it’s understandable that some users would prefer to use `prefer-const` for regular JavaScript outside of that.

(Personally, I always use const consistently, though.😅)

baseballyama:
In this case, there is no actual reassignment, so `prefer-const` would allow it to be declared as `const`. However, semantically, some people may prefer to declare it with `let`.

```js
export const Time3 = () => {
	// If `$state` has object value,
	// Time3 user cn reassign `value.value` even without reassign logic in this composable.
	// Therefore it can be declare as both `const` / `let`
	const value = $state( { value: new Date().toString() });

	return {
		get value() {
			return value;
		}
	}
}
```

baseballyama:
Therefore, I am considering creating a separate rule, `prefer-let`. This rule will enforce the consistent use of `let`.

baseballyama:
@ota-meshi So, what you’re saying is that extending `prefer-const` in the first place is unnecessary?

ota-meshi:
> So, what you’re saying is that extending prefer-const in the first place is unnecessary?

No, I think `svelte/prefer-const` is necessary.
I don't think there's any need to ignore `$state()` from the `svelte/prefer-const` rule check.

Since we don't ignore `svelte/prefer-const` rule checks for mutable objects regardless of `$state()`, I think the rules would be inconsistent if we were to ignore only `$state()` from `svelte/prefer-const` rule checks.

If a primitive value of `$state()` is reported with `prefer-const`, it's just a case of the user using it incorrectly, and I think we can follow it with a different rule (e.g. `prefer-let`).

baseballyama:
> If a primitive value of $state() is reported with prefer-const, it's just a case of the user using it incorrectly, and I think we can follow it with a different rule (e.g. prefer-let).

I still don’t fully understand this part. Since `$state` can hold non-primitive values such as object, are there cases where there is no direct reassignment? In such cases, is it correct for `const` to be enforced? Even if a `prefer-let` rule were implemented, wouldn’t it conflict with `prefer-const`, which enforces `const`?

https://svelte.dev/playground/hello-world?version=5.19.9#H4sIAAAAAAAAE32MzQqDMBAGX2VZCiqIlh6DFXrrO9Qeom4lkMaQrLZFfPcSpT-H0tsy38xOaOSVUOCRtO7h1jvdQkytYmoTTPGiNHkUpwn5YYMXAKav6mBt5kfSHFgtPf3iTW-YDHsUWPjGKctlZSrWxDB4crCHjWfJFMMEoRcQSa0ailKQHQnYbWGGpDJF_qlNYcvVncKPLNxzkdtyWZZsHWT35srYgaFWphWj1APtv1LI_xjhxyJgikx3RsFuoPk8PwFxCBbbPgEAAA==

ota-meshi:
If in that example, `user` should be declared with `let`, then I think `user` should also be declared with `let` in the following source code.
What do you think?

https://eslint-online-playground.netlify.app/#eNq1VNuO0zAQ/RXLQtq2apMCb9nuConLG4J3AlJIJpVL4li2U3VV9d8Z3xK3CQsS2j7U8cw5x3Ozz1TJMv389L5rRceB60QdodFAM7pTpWRCP+ackAY04UUL5IG8UrrQsLgrGlbC3TJ4i33kfLNFe66NvVcg0XE2OEL2XmexDBZCJOhecmt2psvarSqAj0XTQ8TwkVjzNcXoYyQz8midqhvorbhLJNY2C/7t0qEguBGPJoqMnE2Cifm+7FKBPvSgRHDgZ7AzLnpNfjJeZVb9IaKS9BmE0bAAuqagGoZNKjtes31yUNgnhp2TmhwUqWXXkpy+c6D0oHJ6n3Pvd3392vR7xgPSATfCGjcOMcMppGmi5zjbJlCtL+Zo5VwBr58EuMJ5jgXDyYIrqIu+GUk+sYUp+UH5nUoklF3bAq+gst1LkuSGMYuJUw64bzmtm0KnETyn3y3Dj0DNGlAZQeBqla68SE7XmMu4cwwc/YLve2zQF6FZx5E2zJGrTHZVQ08KzilppEUmguXJTF8npUxD+Qfhcb7ND05aFp8wn48nDVy5wzCxSRojLayyt1UYwgiNT4WEGuQGy6mwlSYukLLDGOLb8td6Yv+uS2oNL17VYWqcYZr9P0ZvrtZV9MbwX9EPJy/v8ZqLovyFCijbcbziVsBdJVfytquwP77vOa3g+AGEGWZeMsBQhjOvXwNDxeEH0zl/bngEnnXevhAG+uNtsk22m20Ejt0Tpfl3wyq9niiNk/4HtZlXJcb5J/tCL78B2OY91w==

ota-meshi:
The area I have an issue with is the consistency of the rule.

baseballyama:
I fully understand your perspective.👍

As stated in the documentation, Svelte runes recommend using `let`. Based on my understanding, some users interpret JavaScript rules as being overridden by Svelte rules.
https://svelte.dev/docs/svelte/$state

In your example, since user does not use runes, I understood that standard JavaScript rules apply (without being overridden by Svelte rules). Therefore, I believed that the behavior of this PR was consistent.

Additionally, based on your reasoning, shouldn’t `$props` and `$derived` also consistently use const whenever possible? (I may not yet fully understand this part.)

ota-meshi:
I think the reason why `$derived` should be `let` is because it is changed from outside.
But `$state` doesn't really fit that reason.
For primitive values, the user should assign it somewhere, so it won't be reported to `prefer-const`.
If the reason to use `let` is that the object is mutable, then I think objects with `setters` should use `let`.

>  Svelte runes recommend using `let`.

Hmm... I don't think it explicitly says "recommended". I think it's still a matter of user preference 🤔

baseballyama:
I see. Would adding a config be a good middle ground? Like `{ enforceState: false }`?
I want to accommodate users who wish to use `prefer-const` in some way but also want to consistently use `let` for runes.

ota-meshi:
>  I want to accommodate users who wish to use prefer-const in some way but also want to consistently use let for runes.

If there are users requesting it, I think it makes sense to add the option.

baseballyama:
Got it. In that case, since there are no breaking changes, I will try implementing this option after the v3 release.

baseballyama:
I added `excludedRunes` option.

docs/rules/prefer-const.md
```diff
@@ -46,7 +46,8 @@ This rule reports the same as the base ESLint `prefer-const` rule, except that i
     "error",
     {
       "destructuring": "any",
-      "ignoreReadonly": true
+      "ignoreReadonly": true,
+      "excludedRunes": ["$props", "$state"]
```
ota-meshi, line 50: I think it would be better to explain it as the default.

```suggestion
      "excludedRunes": ["$props", "$derived"]
```

ota-meshi (approved):
LGTM! Thank you!

PR #1069, opened 2025-02-10
feat: added the `require-event-prefix` rule

### LINKED ISSUE #910: Add rule `require-event-prefix`

marekdedic:
Closes #910


packages/eslint-plugin-svelte/src/rules/require-event-prefix.ts
```diff
@@ -0,0 +1,122 @@
+import { createRule } from '../utils/index.js';
+import { type TSTools, getTypeScriptTools } from '../utils/ts-utils/index.js';
+import {
+	type MethodSignature,
+	type Symbol,
+	SymbolFlags,
+	SyntaxKind,
```
ota-meshi, line 7: We don't include typescript as a dependency because we don't want to force users to use it.

Could you please change it to load modules using the functions in packages/eslint-plugin-svelte/src/utils/ts-utils/index.ts other than type-only import?

ota-meshi (changes requested):

packages/eslint-plugin-svelte/src/rules/require-event-prefix.ts
marekdedic: Ok, I extracted them to ts-utils :)

ota-meshi (approved):
LGTM! Thank you!

PR #1113, opened 2025-03-02
Docs: Update require-each-key further reading link

lts20050703:
The current further reading link result in 404 because the tutorial link now has svelte or kit in the url


baseballyama (approved):
Thank you!

PR #1122, opened 2025-03-10
chore: add windows CI

baseballyama:

PR #1124, opened 2025-03-10
fix: support windows path delimiter

### LINKED ISSUE #1120: `svelteKitFileType` parsing broken in Windows

9.22.0

3.0.3

### What did you do?

I came across this issue while debugging the `valid-prop-names-in-kit-pages` rule, which always passed locally on my Windows machine but failed on my Cloudflare-hosted build. At first I thought it was a dependency issue, but I eventually realised the rule wasn't running on windows at all.

I eventually traced this to the `svelteKitFileType` in the `svelteContext` being `null`.

I expected the same rule to fail/pass the same way on my local machine as my Cloudflare build.

### What actually happened?

Could not reproduce the same failure on my local machine.

https://github.com/james-camilleri/eslint-plugin-svelte-repro

### Additional comments

The problem is here: https://github.com/sveltejs/eslint-plugin-svelte/blob/615e9d389ece1b58c5f6a11e7e76cc8e416dfa5a/packages/eslint-plugin-svelte/src/utils/svelte-context.ts#L65

Switching this to `path.sep` seems to handle the issue. Not sure if there are any other repercussions to this though, or if the same issue is present anywhere else in the codebase. Or if the paths should have been normalised elsewhere but weren't.

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1120

I would like to add Windows CI but before that I've created the PR to close #1120.
https://github.com/sveltejs/eslint-plugin-svelte/pull/1122

### DIFF
--- a/packages/eslint-plugin-svelte/src/utils/svelte-context.ts
+++ b/packages/eslint-plugin-svelte/src/utils/svelte-context.ts
@@ -62,7 +62,7 @@ function getSvelteFileType(filePath: string): SvelteContext['svelteFileType'] {
 }
 
 function getSvelteKitFileTypeFromFilePath(filePath: string): SvelteContext['svelteKitFileType'] {
-	const fileName = filePath.split('/').pop();
+	const fileName = filePath.split(/[/\\]/).pop();
 	switch (fileName) {
 		case '+page.svelte': {
 			return '+page.svelte';


james-camilleri:
Thanks!

baseballyama:
I forgot chageset😅

PR #1132, opened 2025-03-17
feat: add `ignorePropertyPatterns` property and rename `ignorePatterns` to `ignoreTypePatterns` in `no-unused-props` rule

baseballyama:
After reconsidering, merging #1135 should also resolve the branded types issue. This is because branded types typically originate from external libraries, meaning their properties are externally defined and should thus be fixed by #1135.

Therefore, this becomes a purely additive feature. However, I still think it’s beneficial to allow ignoring patterns based on both type names and properties.

-> I tested this in a project that uses branded types and confirmed that it’s fixed as expected.


baseballyama:
I’ll merge this before US working hours start to minimize the impact.

Issue #1142, opened 2025-03-18
bug: `no-unused-props` fails with nested object type


AlbertMarashi:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.

### What version of ESLint are you using?

9.22.0

### What version of `eslint-plugin-svelte` are you using?

3.3.1

### What did you do?

```ts
<script lang="ts">
import GenericPopout from "./GenericPopout.svelte"

let {
    position,
}: {
    position: {
        x: number,
        y: number
    },
} = $props()

</script>

<GenericPopout position={position}>
    Test
</GenericPopout>
```

### What did you expect to happen?

No error

### What actually happened?

```ts
  7:5  error  'x' in 'position' is an unused property  svelte/no-unused-props
  7:5  error  'y' in 'position' is an unused property  svelte/no-unused-props
```

### Link to **GitHub Repo** with Minimal Reproducible Example

https://github.com/AlbertMarashi/svelte-eslint-repl

### Additional comments

Version `3.1.0` worked, version `3.2.0` broke it

AlbertMarashi:
`3.3.1` is also broken

AlbertMarashi:
Related:
- https://github.com/sveltejs/eslint-plugin-svelte/issues/1136
- https://github.com/sveltejs/eslint-plugin-svelte/issues/1136
- https://github.com/sveltejs/eslint-plugin-svelte/issues/1134
- https://github.com/sveltejs/eslint-plugin-svelte/issues/1133

baseballyama:
This issue has been fixed in v3.3.2.

PR #1143, opened 2025-03-18
fix: stop reporting child properties in `no-unused-props` when the parent object itself is used

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1142


ota-meshi (approved):
👍

PR #1148, opened 2025-03-20
feat(consistent-selector-style): added support for dynamic classes and IDs

### LINKED ISSUE #1109: `consistent-selector-style`: What to do with dynamic class/ID names?

9

3.0.2

### What did you do?

```svelte
<script lang="ts">
  interface Props {
    level: 1 | 2 | 3;
  }

  let { level }: Props = $props();
</script>

<a class={`level-${level}`}>Click me</a>

<style>
  .level-1 {
    margin-left: 10px;
  }

  .level-2 {
    margin-left: 20px;
  }

  .level-3 {
    margin-left: 30px;
  }
</style
```

I am honestly not sure what should happen. We can't parse dynamic class names, so at this point probably the only thing you can do is give up on checking any class names? (Same thing would happen with IDs). At best, we could try to parse some prefix/suffix from the dynamic class name and only ignore those? @ota-meshi @baseballyama What do you think?

I can imagine adding a rule flag that would switch between giving up and the current behaviour, but I don't really think the current behaviour is useful in many ways (it essentially bans dynamic class/ID names, just in a more stupid way)...

### What actually happened?

3 errors saying I should use a type selector instead

N/A, will make a PR instead once agreed on a solution.

### Additional comments

_No response_

marekdedic:
Closes #1109 

I added parsing of both prefixes and suffixes in cases with dynamic values. If the whole value is dynamic, the rule just gives up for classes/IDs in that component.


ota-meshi (approved):
LGTM! Thanks!

PR #1151, opened 2025-03-24
feat: added the `prefer-svelte-reactivity` rule

marekdedic:


packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-svelte-reactivity/invalid/date01-input.svelte
```diff
@@ -0,0 +1,5 @@
+<script>
+  const variable = new Date(8.64e15);
+</script>
+
+{variable}
```
baseballyama, line 5: I don’t think this needs to be reported. If there’s no reassignment, it shouldn’t need to be reactive.

baseballyama (commented):
I think we need to add tests for .svelte.js files also.

marekdedic:
> I think we need to add tests for .svelte.js files also.

Yes, I wanted to ask about whether that's something we're set up to do, I couldn't find any such tests on the repo...

packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-svelte-reactivity/invalid/date01-input.svelte
marekdedic: It's not about reassignment, it's about internal changes. I'd prefer to be this overly-broad. The alternative is tracking calls to every method that modifies the object (I count 16 on `Date` alone) or writing to any property.

On top of that, once something is exported, it should be reported as well (mostly relevant for `.svelte.ts` files)

I am not sure it's worth it, what do you think?

marekdedic:
@ota-meshi Can you please take a look at the test file? I tried the solution from https://github.com/eslint-community/eslint-utils/issues/249#issuecomment-2747409721, but it didn't work for me :/

ota-meshi:
@marekdedic What have you tried? Have you added `URL` here?

https://github.com/sveltejs/eslint-plugin-svelte/blob/3ddbd83068a8a82ed5bfe638483c7dd7839e681a/packages/eslint-plugin-svelte/tests/utils/utils.ts#L18

marekdedic:
> @marekdedic What have you tried? Have you added `URL` here?

That's exactly what I tried, but it didn't work :(

ota-meshi:
Is it `url01-input.svelte` that you're saying didn't work? I've tried adding `URL: 'readonly', URLSearchParams: 'readonly'` to `globals` in your repository and it seems to work fine.

marekdedic:
Ok, I was adding it in a different place and the value got overriden, thanks :)

README.md
```diff
@@ -219,8 +219,8 @@ export default [
 
 ## Editor Integrations
 
-**Visual Studio Code**  
-Install [dbaeumer.vscode-eslint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint).  
+**Visual Studio Code**
+Install [dbaeumer.vscode-eslint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint).
```
ota-meshi, line 223: Could you please revert the unrelated change? Or add `\` to the end of the line?

```suggestion
**Visual Studio Code**\
Install [dbaeumer.vscode-eslint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint).\
```

packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-svelte-reactivity/invalid/date01-input.svelte
ota-meshi: How about tracking instances and checking mutable function calls?
For example, tracking `new Set()` variables and checking `add()` calls.

For example, I might use:

```js
const ALLOWED_VALUES = new Set(["a", "b"]); // Not reported
function isAllowedValue(v) {
  return ALLOWED_VALUES.has(v);
}

const usedValues ​​= new Set();
function process(v) {
  if (usedValues.has(v)) return;
  usedValues.add(v); // Reported
  // ...
}
```

README.md
marekdedic: This was done by pnpm run update 🤷

ota-meshi: Hmm. Maybe it's because `.editorconfig` was added recently? I don't use the EditorConfig plugin so I'm not sure 😓

ota-meshi: I've changed it to use the `/` notation, so could you rebase it? #1215

packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-svelte-reactivity/invalid/date01-input.svelte
marekdedic: Yeah, that's probably a better way to do it...

packages/eslint-plugin-svelte/src/rules/prefer-svelte-reactivity.ts
```diff
@@ -0,0 +1,243 @@
...
+					[ReferenceTracker.CALL]: true
+				},
+				delete: {
+					[ReferenceTracker.CALL]: true
+				},
+				set: {
+					[ReferenceTracker.CALL]: true
+				}
+			})
+		).length > 0
```
ota-meshi, line 159: I think it can be written as follows:

```suggestion
	    !referenceTracker.iteratePropertyReferences(ctorNode, {
		    clear: {
			    [ReferenceTracker.CALL]: true
		    },
		    delete: {
			    [ReferenceTracker.CALL]: true
		    },
		    set: {
			    [ReferenceTracker.CALL]: true
		    }
	    }).next().done
```

marekdedic: Great, I was struggling a little bit with iterators :D

marekdedic:
I have re-implemented the PR so that it checks all the property calls and only reports if the variable is mutated.

packages/eslint-plugin-svelte/src/utils/ast-utils.ts
```diff
@@ -722,3 +722,16 @@ export function findVariableForReplacement(
...
 }
+
+/**
+ * Check if one node is contained withing the subtree of another node.
+ */
+export function isIn(needle: TSESTree.Node, haystack: TSESTree.Node): boolean {
+	if (needle === haystack) {
+		return true;
+	}
+	if (needle.parent === undefined || needle.parent === null) {
```
ota-meshi, line 733: I prefer to use `==` only in this case.

```suggestion
	if (needle.parent == null) {
```

packages/eslint-plugin-svelte/src/rules/prefer-svelte-reactivity.ts
```diff
@@ -0,0 +1,285 @@
...
+			[ReferenceTracker.READ]: true
+		},
+		search: {
+			[ReferenceTracker.READ]: true
+		},
+		username: {
+			[ReferenceTracker.READ]: true
+		}
+	})) {
+		if (node.parent.type === 'AssignmentExpression') {
```
ota-meshi, line 258: I think we need to check if the node on the left.

```suggestion
		if (node.parent.type === 'AssignmentExpression' && node.parent.left === node) {
```

If we don't check that, I think it will match `x = url.hash`.

ota-meshi (changes requested):
Awesome! Looks almost good to me, but I have two comments.

ota-meshi (approved):
LGTM! Thank you!

ota-meshi:
@baseballyama Just to be sure, could you please check this PR?
This is a new rule that is included in the recommended config.

baseballyama:
I will review it within a few days.👍

marekdedic:
@baseballyama Gentle reminder :)

marekdedic:
No worries, I presume you also do this in your free time :)

PR #1162, opened 2025-03-28
test: account for js output in test cases

43081j:
Currently, we assume test output will always be a `.svelte` file. This isn't always true (e.g. for the derived input/output rule).

This changes the util to basically use the extension of the input in the output, and avoids re-computing it by passing it in.


baseballyama (approved):
Nice! Thank you!

PR #1163, opened 2025-03-28
feat: add fixer for `derived-has-same-inputs-outputs`

43081j:
Simply replaces the param names with the correct ones since we're already aware of what they should be.

Depends on #1162


packages/eslint-plugin-svelte/src/rules/derived-has-same-inputs-outputs.ts
```diff
@@ -53,7 +54,8 @@ export default createRule('derived-has-same-inputs-outputs', {
 					node: fn,
 					loc: fnParam.loc,
 					messageId: 'unexpected',
-					data: { name: expectedName }
+					data: { name: expectedName },
+					fix: (fixer) => fixer.replaceText(fnParam, expectedName)
```
JounQin, line 58: Again, should it be `fix` or `suggestion`?

JounQin (changes requested):

baseballyama:
@43081j 

According to the official ESLint documentation, the following is stated about fixers.

> Best practices for fixes:
> Avoid any fixes that could change the runtime behavior of code and cause it to stop working.
> Make fixes as small as possible. Fixes that are unnecessarily large could conflict with other fixes, and prevent them from being applied.

https://eslint.org/docs/latest/extend/custom-rules#applying-fixes

In addition to this PR, it would also require changes to the places where the argument is used. In that case, we would need to check for any variable name conflicts, otherwise it would violate "Avoid any fixes that could change the runtime behavior of code and cause it to stop working.".

Therefore, implementing this as a fixer doesn’t seem appropriate. Would it be better to make it a suggestion at the very least?

43081j:
similar to the other PR, i am aware of the best practices.

i just overlooked updating references to the parameters since we have no existing tests for those cases

if this becomes more complex than expected, it can be a suggestion.

43081j:
ok @baseballyama i've updated this to rename references of the variable in the current function scope

if you want me to change it into a suggestion once you've re-reviewed, let me know and ill update

packages/eslint-plugin-svelte/src/types.ts
```diff
@@ -366,4 +366,6 @@ export interface SourceCode {
 	getCommentsAfter(nodeOrToken: NodeOrToken | AST.Token): AST.Comment[];
 
 	getCommentsInside(node: NodeOrToken): AST.Comment[];
+
+	getScope(node: NodeOrToken): Scope;
```
43081j, line 370: our types specify that `context.getScope()` exists but it doesn't seem to in tests (it throws if we try use that)

meanwhile, `context.getSourceCode().getScope(node)` does exist

im guessing there's some compat stuff going on here to do with eslint versions removing deprecated stuff

baseballyama: We need to import `getScope` from `ast-utils.js`.

packages/eslint-plugin-svelte/src/rules/derived-has-same-inputs-outputs.ts
```diff
@@ -53,7 +55,20 @@ export default createRule('derived-has-same-inputs-outputs', {
...
+
+						if (!variable) {
+							return fixer.replaceText(fnParam, expectedName);
+						}
+
+						return [
+							fixer.replaceText(fnParam, expectedName),
+							...variable.references.map((ref) => fixer.replaceText(ref.identifier, expectedName))
+						];
+					}
```
baseballyama, line 71: I think it’s not possible to add a fixer because we need to check if `expectedName` is already used.
If `expectedName` is already used, I don’t think we can know how to rename it properly.

43081j: we can check if the name already exists and only add a fixer if it does

baseballyama:
Can you add a changeset?

packages/eslint-plugin-svelte/src/rules/derived-has-same-inputs-outputs.ts
```diff
@@ -1,7 +1,35 @@
...
+	const scope = getScope(context, node);
+	let hasConflict = false;
+	let variable: Variable | null = null;
+
+	for (const v of scope.variables) {
+		if (hasConflict && variable) {
+			break;
+		}
+		if (v.name === expectedName) {
+			hasConflict = true;
```
baseballyama, line 23: I think we need to check parent/global scopes also to check the conflict.

packages/eslint-plugin-svelte/src/rules/derived-has-same-inputs-outputs.ts
```diff
@@ -49,11 +79,35 @@ export default createRule('derived-has-same-inputs-outputs', {
...
+										yield fixer.replaceText(fnParam, expectedName);
+
+										if (variable) {
+											for (const ref of variable.references) {
+												yield fixer.replaceText(ref.identifier, expectedName);
+											}
+										}
+									}
+								}
+							]
```
baseballyama, line 110: Can we merge the logic? There is the same logic below also.

packages/eslint-plugin-svelte/src/rules/derived-has-same-inputs-outputs.ts
43081j: why so?

the param is local to the function body. if the user chose to have a variable named the same in the outer scope, it won't conflict, it'll just cause shadowing (which they should be picking up via the shadowing rule in eslint)

baseballyama: I didn't run tests but I guess...

```js
const $a = 1;
derived(a, (b) => {
  // output 1.
  console.log($a);
});
```

This code will be...

```js
const $a = 1;
derived(a, ($a) => {
  // output a value.
  console.log($a);
});
```

So I think the behavior will change in a way that’s hard to notice and could cause issues.

43081j: you're right. this should be fixed now and i've added a test

packages/eslint-plugin-svelte/src/rules/derived-has-same-inputs-outputs.ts
```diff
@@ -1,7 +1,56 @@
...
+import type { RuleContext, RuleFixer } from '../types.js';
 import { extractStoreReferences } from './reference-helpers/svelte-store.js';
+import { getScope } from '../utils/ast-utils.js';
+
+function findVariableForName(
+	context: RuleContext,
+	node: TSESTree.Node,
+	name: string,
+	expectedName: string
+): { hasConflict: boolean; variable: Variable | null } {
```
baseballyama, line 13: If `hasConflict` is true, we shouldn’t be interested in the `variable` anymore, so I think we could always early return at that point.
(Sorry for the minor comment!)

```suggestion
): { hasConflict: true; variable: null } | { hasConflict: false; variable: Variable | null } {
```

packages/eslint-plugin-svelte/tests/fixtures/rules/derived-has-same-inputs-outputs/invalid/test01-errors.yaml
```diff
@@ -1,32 +1,1044 @@
...
+        derived([e, f], ([g, h]) => {
+        	g;
+        	h;
+
+        	somethingWithACallback(() => {
+        		g;
+        		h;
+        	});
+        });
+        derived([e, f], ([g, $f]) => {
```
baseballyama, line 1015: It should be

```suggestion
        derived([e, f], ([$e, $f]) => {
```

43081j: it shouldn't as the suggestions apply one at a time. this is listing the output of each individual suggestion

baseballyama:
> it shouldn't as the suggestions apply one at a time. this is listing the output of each individual suggestion

Sorry! That's why I deleted the comment but you replied too fast😅

baseballyama (approved):
Thank you!

packages/eslint-plugin-svelte/src/rules/derived-has-same-inputs-outputs.ts
43081j: we can but a variable can still be `null` if there is no conflict, so this type doesn't help us much

things can mutate the AST etc outside of our rule i suppose. `variables` doesn't necessarily contain the variable we're looking for

baseballyama: At the very least, early return might improve performance a bit, but it’s not a big deal.

43081j: i have added it 👍

PR #1164, opened 2025-03-28
feat: add fixer for `no-at-debug-tags`

43081j:
Adds a simple fixer which removes the debug node.


packages/eslint-plugin-svelte/tests/fixtures/rules/no-at-debug-tags/invalid/debug01-output.svelte
```diff
@@ -0,0 +1,11 @@
+<script>
```
JounQin, line 1: There is no `{@debug}` inside, does it been remove by linting fix itself?

JounQin (changes requested):

packages/eslint-plugin-svelte/tests/fixtures/rules/no-at-debug-tags/invalid/debug01-output.svelte
43081j: this is the test output, i.e. after it has passed through the fixer. so it shouldn't be there

generally complex changes are suggestions, and simple ones are fixes. happy to make this a suggestion if people want though

baseballyama:
For example, in my case, I automatically run ESLint fixers with format on save.
If the {@debug} tag gets removed during that process, it effectively makes the {@debug} tag unusable.
Therefore, I believe having a fixer would actually be inconvenient.

I think it would be better to make it a suggestion, similar to the `no-console` rule:
https://eslint.org/docs/latest/rules/no-console

43081j:
I don't mind either way. If your usage is the common use case, I'm happy to change it to a suggestion

baseballyama:
@43081j I would appreciate it if you could make it a suggestion.

43081j:
i've updated it to be a suggestion :+1:

JounQin (changes requested):
CI is broken due to related document unchanged.

baseballyama:
A you add a changeset?

baseballyama (approved):
Thank you!

PR #1165, opened 2025-03-28
feat: add fixer to `require-store-callbacks-use-set-param`

43081j:
Adds a fixer which adds or changes the param to `set`.

For example:

```ts
// before
readable(val, () => ...);
readable(val, function () { ... });

// after
readable(val, (set) => ...);
readable(val, function (set) { ... });
```


packages/eslint-plugin-svelte/src/rules/require-store-callbacks-use-set-param.ts
```diff
@@ -27,7 +28,20 @@ export default createRule('require-store-callbacks-use-set-param', {
 						context.report({
 							node: fn,
 							loc: fn.loc,
-							messageId: 'unexpected'
+							messageId: 'unexpected',
+							fix: (fixer) => {
+								if (param) {
+									return fixer.replaceText(param, 'set');
+								}
```
baseballyama, line 35: Renaming a parameter would break the parts of the code that use that variable. This doesn’t seem to align with ESLint’s fixer policy.

https://eslint.org/docs/latest/extend/custom-rules#applying-fixes

> Best practices for fixes:
> Avoid any fixes that could change the runtime behavior of code and cause it to stop working.
> Make fixes as small as possible. Fixes that are unnecessarily large could conflict with other fixes, and prevent them from being applied.

43081j: You're right. It does certainly align with the best practices, I just overlooked this part. We should be updating references to the parameter with the new name

packages/eslint-plugin-svelte/src/rules/require-store-callbacks-use-set-param.ts
```diff
@@ -1,5 +1,36 @@
+import { getScope } from '../utils/ast-utils.js';
+import type { RuleContext, SuggestionReportDescriptor } from '../types.js';
 import { createRule } from '../utils/index.js';
 import { extractStoreReferences } from './reference-helpers/svelte-store.js';
+import type { TSESTree } from '@typescript-eslint/types';
+import type { Variable } from '@typescript-eslint/scope-manager';
+import { getSourceCode } from '../utils/compat.js';
+
+function findVariableForName(
```
43081j, line 9: this is a duplicate of #1163 so i'll move it into a util once that one lands and update this

baseballyama (approved):

PR #1173, opened 2025-03-31
chore: use `context.sourceCode` instead of compat

43081j:
As of 3.x, we only support `8.57.1` and `>=9.0.0` of ESLint.

Both of these have `context.sourceCode` available, so we should no longer need the `getSourceCode` compat helper.

The helper does polyfill some missing methods but `8.57.1` seems to have all of those already.

### DIFF
--- a/.changeset/happy-donuts-fold.md
+++ b/.changeset/happy-donuts-fold.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': patch
+---
+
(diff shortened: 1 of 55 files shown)


baseballyama (commented):
Can you please update `packages/eslint-plugin-svelte/eslint.config.mjs` line number 73?

43081j:
i had the same thought and just made the change as you wrote that, in the other PR too 👍

baseballyama (approved):
Thank you!

PR #1174, opened 2025-03-31
chore: use `context.filename` and `context.physicalFilename`

43081j:
Since we now require latest 8.x, all versions of ESLint we support have both `context.filename` and `context.physicalFilename`.

In the old types, `physicalFilename` could be undefined but now will actually default to `filename` (so the type has also changed).

If #1173 lands first, this will have conflicts i'll resolve before we can merge.

### DIFF
--- a/.changeset/petite-impalas-reply.md
+++ b/.changeset/petite-impalas-reply.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': patch
+---
+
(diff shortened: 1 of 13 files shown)


43081j:
i'll open a 3rd PR to remove the file and the dependency entirely once this and #1173 are merged

since we can do the same for `cwd` too

PR #1176, opened 2025-03-31
chore: use `context.cwd`

43081j:
Switches to using `context.cwd` since all ESLint versions we support now have it.

The previously used `getCwd` helper falls back to `process.cwd()` (which ESLint itself doesn't do). So in the few places we relied on that, a fall back to `process.cwd()` has been added.

~This is the last remaining compat function, so we have also dropped the `eslint-compat-utils` dependency.~


43081j:
weirdly this CI failure doesn't seem related but im not sure why else it'd be happening 🤔

43081j:
it was me being blind. we still depend on the compat package for the rule tester

baseballyama:
@43081j Will you make a 1 more PR for `packages/eslint-plugin-svelte/tests/utils/eslint-compat.ts`?

43081j:
i was looking at that, i feel like we should be able to drop most of it (or all of it?)

i'll see where i get with it 👍

PR #1177, opened 2025-03-31
test: use `Linter` and `ESLint` directly from eslint in tests

43081j:
This drops a few of the compat exports in tests and imports directly from eslint instead.

`Linter` already seems to work in our tests for 8.x and 9.x directly.

`ESLint` is the flat class in 9.x, but the legacy one in 8.x. Luckily, it provides `loadESLint` which can retrieve the flat `ESLint` for us in both versions, so we should no longer need the compat helper.

We can't quite get rid of the `RuleTester` compat layer as 8.x doesn't export the flat rule tester from the index (it is only available through the unsupported-api entrypoint).

### DIFF
--- a/.changeset/hungry-days-reply.md
+++ b/.changeset/hungry-days-reply.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': patch
+---
+
+test: use ESLint and Linter directly from eslint in tests


packages/eslint-plugin-svelte/tests/src/configs/all.ts
```diff
@@ -1,16 +1,20 @@
 import assert from 'assert';
-import semver from 'semver';
 import plugin from '../../../src/index.js';
-import { ESLint } from '../../utils/eslint-compat.js';
+import { loadESLint, type ESLint as ESLintClass } from 'eslint';
```
43081j, line 3: as far as i understand, `ESLint` here is the flat ESLint already. but when we force install 8.x in CI, it will be the legacy one at runtime. which is why we can't just `import {ESLint} from 'eslint'`

PR #1185, opened 2025-04-02
docs: update valid-each-key, require-each-key, no-at-html-tags, prefer-class-directive further reading links

andrewpollack:
These currently lead to 404s. Updating links as well as described paths to match current tutorial


ota-meshi (approved):
Thank you!

PR #1187, opened 2025-04-03
chore: move compat dependency to devDependencies

43081j:
We no longer use it in production code, only in tests. So we can move this to `devDependencies` now.

### DIFF
--- a/packages/eslint-plugin-svelte/package.json
+++ b/packages/eslint-plugin-svelte/package.json
@@ -84,7 +84,6 @@
     "acorn": "^8.14.0",
     "assert": "^2.1.0",
     "esbuild": "^0.25.0",
-    "eslint-compat-utils": "^0.6.4",
     "eslint-scope": "^8.2.0",
     "eslint-typegen": "^2.0.0",
     "eslint-visitor-keys": "^4.2.0",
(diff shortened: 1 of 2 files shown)


JounQin:
Is that really needed? We requires `"eslint": "^8.57.1 || ^9.0.0"`.

43081j:
from what i remember, in 8.57.1, the _flat_ rule tester is in `eslint/use-at-your-own-risk`

the index exports the legacy rule tester

in 9.x, the index exports the flat rule tester, and `use-at-your-own-risk` doesn't export one at all

so we'd need a conditional import of some sort, or continue using the compat utils

JounQin:
I personally think that's easy enough to use `eslint.RuleTester || eslintUnsupportedApi.FlatRuleTester` other than depending on a dep we don't use much.

43081j:
ok i've pushed a change

it seems to work locally with eslint 8. please take a look

baseballyama (approved):
I think it’s fine, but let’s have @ota-meshi, the owner of `eslint-compat-utils`, confirm it as well.

43081j:
when reviewing it, @ota-meshi, keep in mind this is only in tests.

your compat code does some patching up of `FlatRuleTester` but it doesn't seem we need that here

ota-meshi (approved):
It's great that eslint-compat-utils is no longer needed! Thanks!

ota-meshi:
Oh... We should have included the changeset since this is a dependency removal.

Issue #1194, opened 2025-04-14
new rule: `svelte/no-add-event-listner`


baseballyama:
### Motivation

We need to use `on` instead.
https://svelte.dev/docs/svelte/svelte-events#on

### Description

Since Svelte 5 uses event delegation, using `addEventListener` may cause events to fire in unexpected order.
Therefore, we should always use `on` instead.

### Examples

```svelte
<script>
  import { on } from 'svelte/events';

  onMount(() => {
    // ✓ GOOD
    on(parent, 'click',() => {});
    // ✗ BAD
    parent.addEventListener('click',() => {});
  })
</script>

<div bind:this={parent}>
  <button />
</div>
```

PR #1197, opened 2025-04-17
feat: add `no-add-event-listener` rule

43081j:
this helped me learn about the events helpers, so figured i may as well throw it together.

let me know if there's anything i missed, as i don't think i've added a new rule in this repo before.

one thing to note is that it has a suggestion, but that relies on the user later making sure `on` has been imported correctly. im not sure we can do that automatically since each suggestion runs individually and should probably only affect the subset of code you're reporting

close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1194


docs/rules/no-add-event-listener.md
```diff
@@ -0,0 +1,31 @@
...
+This rule reports usages of `addEventListener`:
+
+<!--eslint-skip-->
+
+```svelte
+<script>
+  /* eslint svelte/no-add-event-listener: "error" */
+  addEventListener(window, 'resize', handler);
+</script>
+```
```
baseballyama, line 26: Can you add OK case also?

packages/eslint-plugin-svelte/src/rules/no-add-event-listener.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import { createRule } from '../utils/index.js';
+import type { SuggestionReportDescriptor } from '../types.js';
+
+export default createRule('no-add-event-listener', {
+	meta: {
+		docs: {
+			description: 'Warns against the use of `addEventListener`',
+			category: 'Best Practices',
+			recommended: false,
```
baseballyama, line 11: Do you think this should be shipped as a recommended rule? (It might cause a lot of errors in existing code, though.)
cc: @ota-meshi

packages/eslint-plugin-svelte/tests/fixtures/rules/no-add-event-listener/invalid/test01-input.svelte
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	addEventListener('message', handler);
+	window.addEventListener('message', handler);
+	document.addEventListener('visibilitychange', handler);
+
+	// with a load of whitespace
+	window.addEventListener    ('message', handler);
+	// with a comment
+	window.addEventListener/* foo */('message', handler);
```
baseballyama, line 28: Should we also include the TypeScript case like this?

```ts
(window as any).addEventListener ('message', handler);
(window.addEventListener as any)('message', handler);
```

packages/eslint-plugin-svelte/tests/fixtures/rules/no-add-event-listener/invalid/test01-input.svelte
```diff
@@ -0,0 +1,31 @@
...
+			}
+		};
+
+		target.addEventListener('focus', handler);
+		deepObj.deep.obj.target.addEventListener('focus', handler);
+	}
+
+	addEventListener('message', handler);
+	window.addEventListener('message', handler);
+	document.addEventListener('visibilitychange', handler);
```
baseballyama, line 23: Do we need to add tests for 3rd argument?
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

baseballyama (commented):
I’d really appreciate it if we could help by adding some tests for now!

packages/eslint-plugin-svelte/src/rules/no-add-event-listener.ts
43081j: ultimately i think it should be `recommended`, but im not sure how we can do that without introducing loads of errors like you say 🤔 it'd have to be in the next major i suppose

packages/eslint-plugin-svelte/tests/fixtures/rules/no-add-event-listener/invalid/test01-input.svelte
43081j: for this one, i added the test cases but this one isn't currently supported:

```ts
(window.addEventListener as any)('message', handler);
```

we would need logic to handle call expressions containing casts, which feels weird. maybe we should just ignore this case and not try handle it?

packages/eslint-plugin-svelte/src/rules/no-add-event-listener.ts
baseballyama: Let’s make it recommended in the next major release.

docs/rules/no-add-event-listener.md
```diff
@@ -0,0 +1,42 @@
...
+# svelte/no-add-event-listener
+
+> Warns against the use of `addEventListener`
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> **_This rule has not been released yet._** </badge>
+- :bulb: Some problems reported by this rule are manually fixable by editor [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
+
+## :book: Rule Details
+
+This rule reports usages of `addEventListener`:
```
baseballyama, line 17: Is it better?

```suggestion
Svelte relies on event delegation for performance and predictable handler order. Calling `addEventListener` inside a component skips this mechanism. This rule reports any call to `addEventListener` suggests converting to the `on()` helper from `svelte/events`.
```

packages/eslint-plugin-svelte/src/rules/no-add-event-listener.ts
```diff
@@ -0,0 +1,74 @@
...
+			default: 'warn'
```
baseballyama, line 12: Why was the default set to warn?

packages/eslint-plugin-svelte/tests/fixtures/rules/no-add-event-listener/invalid/test01-input.svelte
baseballyama: I’m not sure how complicated this is, so I’ll try implementing it myself later.

packages/eslint-plugin-svelte/src/rules/no-add-event-listener.ts
```diff
@@ -0,0 +1,74 @@
...
+			}
+		]
+	},
+	create(context) {
+		return {
+			CallExpression(node: TSESTree.CallExpression) {
+				const { callee, arguments: args } = node;
+				let target: string | null = null;
+
+				if (args.length < 2 || args.length > 3) {
```
baseballyama, line 33: Can we check such weird pattern? 😅

```svelte
<script>
	const foo = ['resize', () => { console.log('resized!') }]
	window.addEventListner(...foo);
</script>
```

baseballyama: Even if suggestions aren’t possible, maybe it’s fine to just report any use of `addEventListener`.

baseballyama (commented):
Please add changeset🙏

packages/eslint-plugin-svelte/tests/fixtures/rules/no-add-event-listener/invalid/test01-input.svelte
43081j: its not complicated, its just awkward that we'd need code which knows about casts

```ts
((window.addEventListener) as any) as any)('message', handler)
```

the callee is then a `TSAsExpression`

which means we'd need to traverse this node specifically stripping out/ignoring casts (because only casts create this situation)

i.e. you would do this flow:

```
const actualNode = findFirstNonCastNode(node);

// that actually does this
while (expression.type is TSAsExpression) {
  expression = expression.expression;
}
```

sure it'd make it work but it makes me feel like this for some reason: 😬

docs/rules/no-add-event-listener.md
43081j: i think so 👍  thanks for rewording

packages/eslint-plugin-svelte/src/rules/no-add-event-listener.ts
43081j: you're probably right and we don't need to care how many params there are, or what they are

i'll add a test for spread, to make sure we don't provide suggestions for it

packages/eslint-plugin-svelte/src/rules/no-add-event-listener.ts
43081j: some rules set this and some don't. i was trying to be consistent with the other ones, but have removed it since most don't specify this

packages/eslint-plugin-svelte/src/rules/no-add-event-listener.ts
ota-meshi: We don't have a mechanism for that yet, but we may be able to include this rule in the recommended set as a warning.

ota-meshi: Of course we can work on this PR separately.

43081j:
@baseballyama any chance you can take one last look at this? it should be good to go I think

ota-meshi (approved):
Looks good to me! Thanks!

Issue #1180, opened 2025-04-01
svelte/no-unnecessary-state-wrap cause unsolvable error with svelte(non_reactive_update)


gyzerok:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.

### What version of ESLint are you using?

9.23.0

### What version of `eslint-plugin-svelte` are you using?

3.4.1

### What did you do?

In the following code snippet if I remove `$state` rune then I get `svelte(non_reactive_update)` error, however if I add it I get `svelte/no-unnecessary-state-wrap`.

<details>
<summary>Configuration</summary>

```
export default ts.config(
	includeIgnoreFile(gitignorePath),
	js.configs.recommended,
	...ts.configs.recommended,
	...svelte.configs.recommended,
	prettier,
	...svelte.configs.prettier,
	{
		languageOptions: {
			globals: { ...globals.browser, ...globals.node }
		},
		rules: { 'no-undef': 'off' }
	},
	{
		files: ['**/*.svelte', '**/*.svelte.ts', '**/*.svelte.js'],
		ignores: ['eslint.config.js', 'svelte.config.js'],
		languageOptions: {
			parserOptions: {
				projectService: true,
				extraFileExtensions: ['.svelte'],
				parser: ts.parser,
				svelteConfig
			}
		}
	}
);
```
</details>

```svelte
<script lang="ts">
	import { SvelteSet } from 'svelte/reactivity';
	import Bug3 from './Bug3.svelte';

	let set = new SvelteSet<number>([]);

	$effect(() => {
		console.log(set);
	});
</script>

<Bug3 bind:set />

```

### What did you expect to happen?

Not entirely sure

### What actually happened?

Either of these errors
```
❯ npm run lint

> svelte-eslint-bug@0.0.1 lint
> prettier --check . && eslint .

Checking formatting...
All matched files use Prettier code style!

svelte-eslint-bug/src/routes/+page.svelte
  5:19  error  SvelteSet is already reactive, $state wrapping is unnecessary  svelte/no-unnecessary-state-wrap

✖ 1 problem (1 error, 0 warnings)
```
```
❯ npm run check

> svelte-eslint-bug@0.0.1 check
> svelte-kit sync && svelte-check --tsconfig ./tsconfig.json


====================================
Loading svelte-check in workspace: svelte-eslint-bug
[...]

Issue #1154, opened 2025-03-25
no-unnecessary-state-wrap triggers too often


benmccann:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.

### What version of ESLint are you using?

9

### What version of `eslint-plugin-svelte` are you using?

3.3.3

### What did you do?

Tried to upgrade https://github.com/immich-app/immich/tree/main/web to 3.3.3

It triggered `no-unnecessary-state-wrap` in two places. One was helpful, but the other I consider to be a false positive

### What did you expect to happen?

This shouldn't be `$state` and should be a `const`: https://github.com/immich-app/immich/blob/392ce7deb2683d2c66a821c617329b3641491016/web/src/lib/components/forms/tag-asset-form.svelte#L21

However, this one is pretty reasonable because you're creating a new `SvelteSet` and it's a lot more cumbersome to `clear` the set and then `add` each item: https://github.com/immich-app/immich/blob/90f21d9047aa33dcb8231ab61269e60a32aedd9f/web/src/lib/components/utilities-page/duplicates/duplicates-compare-control.svelte#L86

### What actually happened?

It said that I should not use `$state` in either location.

### Link to **GitHub Repo** with Minimal Reproducible Example

The full repo is here: https://github.com/immich-app/immich/tree/main/web

I can create a smaller version if needed

baseballyama:
We can suppress the warning by setting `allowReassign` to true.
https://sveltejs.github.io/eslint-plugin-svelte/rules/no-unnecessary-state-wrap/#options

I think we can discuss the default value. Personally, I slightly preferred using `clear` & `add` to keep the reactive tree smaller.

marekdedic:
@benmccann I originally thought the same and argued with @baseballyama in #1062 but in the end, I agree that assigning to `svelte/reactivity` type variables is mostly an anti-pattern. Maybe the docs could be better for this rule, however?

benmccann:
Catching up on that thread, I agree that `map = new SvelteMap();` could be considered an anti-pattern as you can simply do `map.clear()`.

However, in the case that I ran into, the code was `selectedAssetIds = new SvelteSet(assets.map((asset) => asset.id));`. This does not seem like an anti-pattern to me as it's far more cumbersome to write:
```
selectedAssetIds.clear();
[...]

PR #1202, opened 2025-04-19
feat(no-unnecessary-state-wrap): support string array in allowReassign, default to [‘SvelteSet’], and improve messages

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1154
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1180

### DIFF
--- a/docs/rules/no-unnecessary-state-wrap.md
+++ b/docs/rules/no-unnecessary-state-wrap.md
@@ -66,14 +66,17 @@ Therefore, wrapping them with `$state` is unnecessary and can lead to confusion.
     "error",
     {
       "additionalReactiveClasses": [],
-      "allowReassign": false
+      "allowReassign": ["SvelteSet"]
     }
   ]
 }
 ```
 
 - `additionalReactiveClasses` ... An array of class names that should also be considered reactive. This is useful when you have custom classes that are inherently reactive. Default is `[]`.
-- `allowReassign` ... If `true`, allows `$state` wrapping of reactive classes when the variable is reassigned. Default is `false`.
+- `allowReassign` ... Can be either a boolean or an array of class names:
+  - If `true`, allows `$state` wrapping of any reactive classes when the variable is reassigned.
+  - If an array, allows `$state` wrapping for the specified reactive classes when the variable is reassigned.
+  - Default is `["SvelteSet"]`.
 
 ### Examples with Options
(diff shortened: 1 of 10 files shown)


benmccann:
I think `set = new SvelteSet([1, 2, 3])` should be allowed as it's cumbersome to write another way, but `set = new SvelteSet()` should not be because `set.clear();` is easier to write

baseballyama:
> I think set = new SvelteSet([1, 2, 3]) should be allowed as it's cumbersome to write another way, but set = new SvelteSet() should not be because set.clear(); is easier to write

Since the rule would become more complex, would it be better to implement it as a separate rule, like `prefer-svelteset-new`?

marekdedic:
Why should `SvelteSet` be allowed by default? I don't see why it's different from e.g. `SvelteMap`, am I missing something?

benmccann:
> Since the rule would become more complex, would it be better to implement it as a separate rule, like prefer-svelteset-new?

I think that would be fine. The name seems backwards to me though. Perhaps something like `prefer-clear-over-reassignment`. I also think you don't necessarily need `svelteset` in the name because it could apply to `SvelteMap` as well

> Why should SvelteSet be allowed by default? I don't see why it's different from e.g. SvelteMap, am I missing something?

I agree. I think we ended up deciding all reassignments should be allowed by default though, so probably not an issue once that change is made. The `prefer-clear-over-reassignment` rule can handle the `SvelteSet` and `SvelteMap` cases

marekdedic:
Thanks 🙏 

I'm a bit lost and want to understand this: Am I correct that what this PR essentially does is by default turn off the rule for any SvelteSet or SvelteMap that is re-assigned anywhere?

baseballyama:
> I'm a bit lost and want to understand this: Am I correct that what this PR essentially does is by default turn off the rule for any SvelteSet or SvelteMap that is re-assigned anywhere?

The purpose of this PR is to change the default value of `allowReassign` to `true`, and to exclude any reassigned reactive classes including `SvelteSet` and `SvelteMap` from being reported.

baseballyama:
I’ve updated the PR title, and I’ll update the PR this weekend.

benmccann:
@baseballyama just a friendly reminder that this PR is still in draft state

PR #1235, opened 2025-05-20
feat: support comma-separated ignore comments

43081j:
This is two things:

- performance improvements to the algorithm we use for extracting ignore comments
- support for comma-separated ignore comments

We can now support (as per svelte docs):

```html
<!-- svelte-ignore code1, code2 -->
```

It also drops multi-line matching in the pattern to avoid matching this kind of funk:

```html
<!-- some
random
multi-line
content
svelte-ignore code1
more content
-->
```


packages/eslint-plugin-svelte/tests/fixtures/rules/no-unused-svelte-ignore/valid/svelte-ignore-comma-separated-input.svelte
```diff
@@ -0,0 +1,5 @@
+<!-- svelte-ignore a11y_label_has_associated_control, a11y_no_noninteractive_tabindex -->
```
ota-meshi, line 1: To exclude Svelte v4/v3 tests, add a `svelte-ignore-comma-separated-requirements.json` file and write the following:

```json
{
	"svelte": ">=5.0.0"
}
```

.changeset/quick-cloths-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-svelte': patch
```
ota-meshi, line 2:
```suggestion
'eslint-plugin-svelte': minor
```

ota-meshi (changes requested):
Thank you for this PR!
I have a change request to resolve some old Svelte test failures.

packages/eslint-plugin-svelte/tests/fixtures/rules/no-unused-svelte-ignore/valid/svelte-ignore-comma-separated-input.svelte
ota-meshi: By the way, the reason for the failure is that the error message format is different in old Svelte 😅

43081j: i did add a requirements file, the one above this comment

but maybe i put it in the wrong place or something 🤔

PR #1244, opened 2025-06-08
test: added support for .svelte.js test files

marekdedic:
I have already tested the changes in #1151, but decided to put them into a separate PR and they can be merged in any order :)


ota-meshi (approved):
Thank you!

PR #1272, opened 2025-07-17
chore: warn unnecessary-state-wrap

hyunbinseo:
There are cases where `SvelteSet` has to be reassigned. Reference https://github.com/sveltejs/svelte/issues/16422#issuecomment-3084031841

```svelte
<script lang="ts">
  import { SvelteSet } from 'svelte/reactivity';

  // SvelteSet is already reactive, $state wrapping is unnecessary.
  let ids = $state(new SvelteSet([1]));
  const count = $derived(ids.size); // added
</script>

{#each ids as id (id)}{id}{/each}

<button type="button" onclick={() => (ids = new SvelteSet([2]))}>
  Update IDs
</button>
```

Since it is considered a [best practice](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-unnecessary-state-wrap/), I believe the recommendation can be lowered from error to warn.

This will allow devs to comment out the ESLint warning when needed with confidence.

### DIFF
--- a/.changeset/gentle-buses-trade.md
+++ b/.changeset/gentle-buses-trade.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': patch
+---
+
+chore: warn unnecessary-state-wrap
--- a/packages/eslint-plugin-svelte/src/configs/flat/recommended.ts
+++ b/packages/eslint-plugin-svelte/src/configs/flat/recommended.ts
@@ -32,7 +32,7 @@ const config: Linter.Config[] = [
 			'svelte/no-store-async': 'error',
 			'svelte/no-svelte-internal': 'error',
 			'svelte/no-unknown-style-directive-property': 'error',
-			'svelte/no-unnecessary-state-wrap': 'error',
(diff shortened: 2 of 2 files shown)


marekdedic:
Hi,
this is a little bit more complicated issue and I don't think there is a clear agreement :/ For my 2 cents, in your example, the rule pushes you towards using `Set.clear()`, which is IMO better that re-assignening (at least performance-wise)

hyunbinseo:
@marekdedic Thanks for the comment.

I am trying to clear and re-add multiple (100+) new items from an array.

Is `set.clear()` and `arr.forEach((v) => set.add(v))` still the advised method?

marekdedic:
@hyunbinseo In the end, you just have to measure the variants to truly know performance. My intuition would be that minimising updates would be the best option, but Svelte today is pretty good at optimising things. So I'd prefer modifying items over re-adding over recreating the whole set. But it depends on your application in the end...

ota-meshi:
If you want that rule to be a warning, feel free to change it in your configuration :smiley:

PR #1273, opened 2025-07-23
fix: add `ignoreLocalVariables` to `prefer-svelte-reactivity` rule to avoid false positives

baseballyama:


marekdedic:
Thanks for taking this on. I think we should be careful, though - this should still report even when ignoring local variables:

```svelte.ts
function fn(): Set<string> {
  const local = new Set();
  local.add("Hello");
  return local;
}
```

Arguably, even this should report (and probably doesn't even currently...):


```svelte.ts
function fn(): Set<string> {
  const local = new Set();
  return local;
}
```

(similarly for class members...)

marekdedic:
I took another approach in #1287 which I think is more precise but got inspired by your PR in some places, thanks!

PR #1287, opened 2025-08-06
feat(prefer-svelte-reactivity): ignoring encapsulated local variables by default

### LINKED ISSUE #1271: svelte/prefer-svelte-reactivity triggers inside non-reactive functions (e.g., URL, Date, Set usage)
Version: `eslint-plugin-svelte@3.11.0`

The `svelte/prefer-svelte-reactivity` rule currently triggers even inside regular functions where Svelte's reactivity is not needed or doesn't make sense.

e.g.:

```ts
const getNextPage = () => {
  const newUrl = new URL(page.url) // !!! This leads to an error
  newUrl.pathname = page.url.pathname
  newUrl.searchParams.set('p', (pageNum + 1).toString())
  return newUrl.toString()
}
```
In this context:

- `new URL(...)` is a straightforward operation.
- There's no reactive dependency tracking needed.
- Restructuring this logic into a reactive statement (e.g., using SvelteURL()) would be unnecessary.
- I would also expect the reactive version to be slower than the native one.

I've encountered similar false positives with other built-ins like Date and Set.

marekdedic:
Fixes #1271


marekdedic:
Hey @ota-meshi @baseballyama, sorry to ping you, but could you please help me with the CI? I have no idea why it fails and how to debug it :(

marekdedic:
Also, I'll be away for the next 2 weeks, so I won't be able to respond or fix anything. But I've enabled editing by maintainers, so if you want, feel free to edit the PR and merge it while I'm away :)

ota-meshi (changes requested):
Sorry for the late reply.
I think this PR contains changes for multiple purposes.
Could you split it into several PRs?
I think at least the following two should be split:

- Add the `ignoreEncapsulatedLocalVariables` option
- Tracking private properties

marekdedic:
Ok, I've started in #1326 but I'm afraid the PRs will have to be (mostly) sequential because otherwise there'll just be too many conflicts.

mikededo:
@marekdedic Is there any plan on updating/fixing this? Any constructor call within a non-reactive closure triggers false positives.

marekdedic:
@mikededo I plan on going back over `prefer-svelte-reactivity`, but currently, I'm preoccupied with `no-navigation-without-resolve` (see #1481), which is higher priority since it affects more people :/

PR #1299, opened 2025-08-22
feat: limiting more rules to prevent FPs

### LINKED ISSUE #1279: Disable the `svelte/experimental-require-strict-events` rule with rune components
When I enable the rule in `eslint.config.js`, like this if you take the default `npx sv create` project :
```js
import { includeIgnoreFile } from '@eslint/compat';
import js from '@eslint/js';
import prettier from 'eslint-config-prettier';
import svelte from 'eslint-plugin-svelte';
import globals from 'globals';
import { fileURLToPath } from 'node:url';
import ts from 'typescript-eslint';
import svelteConfig from './svelte.config.js';

const gitignorePath = fileURLToPath(new URL('./.gitignore', import.meta.url));

export default ts.config(
	includeIgnoreFile(gitignorePath),
	js.configs.recommended,
	ts.configs.recommended,
	svelte.configs.recommended,
	prettier,
	svelte.configs.prettier,
	{
		languageOptions: {
			globals: { ...globals.browser, ...globals.node }
		},
		rules: {
			'no-undef': 'off'
		}
	},
	{
		files: ['**/*.svelte', '**/*.svelte.ts', '**/*.svelte.js'],
		languageOptions: {
			parserOptions: {
				projectService: true,
				extraFileExtensions: ['.svelte'],
				parser: ts.parser,
				svelteConfig
			}
		}
	},
	{
		rules: {
			'svelte/experimental-require-strict-events': 'error'
		}
	}
);
```
 My runes components show errors:

Image

I have both legacy and runes components in my project, but I only want the rule to apply to the legacy components.

I'm using the latest `eslint-plugin-svelte` `3.11.0` and `svelte` `5.0.0`

marekdedic:
- **feat: disabling more rules in runes mode**
- **feat: restricting SvelteKit rules to SvelteKit**

Fixes #1279


packages/eslint-plugin-svelte/src/rules/no-dupe-on-directives.ts
```diff
@@ -15,7 +15,16 @@ export default createRule('no-dupe-on-directives', {
...
 		},
-		type: 'problem'
+		type: 'problem',
+		conditions: [
+			{
+				svelteVersions: ['3/4']
+			},
+			{
+				svelteVersions: ['5'],
+				runes: [false, 'undetermined']
```
ota-meshi, line 25: `on:` is deprecated but I think it can still be used, so I think we need the rule to work in rune mode too.

ota-meshi (changes requested):
Thank you for this PR!

ota-meshi (approved):
LGTM! Thank you!

baseballyama (approved):
Thank you!

PR #1308, opened 2025-09-03
feat(no-navigation-without-resolve): added to recommended rule set

marekdedic:
Since the rule is now restricted to running only on SvelteKit files, I think it's safe to add it to the recommended rule set.

### DIFF
--- a/.changeset/bitter-aliens-relax.md
+++ b/.changeset/bitter-aliens-relax.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': minor
+---
+
+feat(no-navigation-without-resolve): added to recommended rule set
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ These rules relate to SvelteKit and its best Practices.
 | Rule ID | Description |    |
 |:--------|:------------|:---|
 | [svelte/no-export-load-in-svelte-module-in-kit-pages](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-export-load-in-svelte-module-in-kit-pages/) | disallow exporting load functions in `*.svelte` module in SvelteKit page components. | :star: |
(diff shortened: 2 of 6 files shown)


marekdedic:
Wow, that was fast :D

DePasqualeOrg:
After updating to the latest version, I'm getting errors that I'm not sure how to resolve. What should I do about `mailto` URLs? It's also showing an error when I use a ternary expression with both branches using `resolve`.

And I believe this should work, but I'm getting an error on `resolve`:

```svelte
<a href={resolve('/blog/[slug]', { slug })}>
```
`Error: Expected 1 arguments, but got 2.ts(2554)`

marekdedic:
Hello, thanks for the comments:
- `mailto:` URLs were an oversight and they should be fixed in version 3.12.1
- In the ternary, we could add that as an exception, but it seems to me to be actually nudging you towards a better design, which is to resolve as the last operation (i.e. do `resolve(condition ? "/a" : "/b")` instead of `condition ? resolve("/a") : resolve("/b")`)
- As for the 2 arguments: Is that an error with this plugin? Or is it a type issue (and therefore more likely an issue with SvelteKit's types)? Can you send a playground link?

DePasqualeOrg:
I have a ternary expression in which one branch is a `mailto` URL and the other uses `resolve`. I'm guessing that will still result in an error after `mailto` URLs are allowed.

The type error regarding arguments might be due to the fact that I have a page route at `/blog` that is not linked to by any pages. The Svelte files for the route exist, but maybe `resolve` uses a list of routes that are discovered through links to determine which are valid?

marekdedic:
Hmm, that's an edge case if I've ever seen one :D There's already a few PRs that would conflict with this, so give us some time please.

hyunbinseo:
If this is a recommended rule, shouldn't the SvelteKit tutorial, docs, and default templates be updated as well?

```svelte
<!-- https://svelte.dev/docs/kit/load#Layout-data -->
<a href="/blog/{post.slug}">
```

image

marekdedic:
Hey @DePasqualeOrg I split the one hanging issue into #1389, but just a heads-up, it's mnot high on my priority list. Otherwise I consider this conversation resolved

PR #1310, opened 2025-09-03
Fix spelling of 'internal' and 'function'

bz2:
In docs for navigation lint rules.


bz2:
@JounQin Thank you for the review - I added a change that fixes 'shallow outing' in the same two doc files which I noticed after putting this PR up.

marekdedic:
Thanks!

Issue #1319, opened 2025-09-04
svelte/no-navigation-without-resolve does not allow `ResolvedPathname` as href


AndersRobstad:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.

### What version of ESLint are you using?

9.34.0

### What version of `eslint-plugin-svelte` are you using?

3.12.1

### What did you do?

<details>

No real config for this rule used, its included in the svelte recommended rules

</details>

```svelte
<script lang="ts">
	import { type ResolvedPathname } from '$app/types';

	interface Props {
		href: ResolvedPathname;
	}

	const { href }: Props = $props();
</script>

<!-- This shows error `Found a link with a url that isn't resolved.` -->
<a href={href}>
     Link here to a resolved pathname
</a>

```

### What did you expect to happen?

I would expect to be able to pass in href's into a link which has types that imply that they are resolved, and thus valid hrefs for the application. This is very useful when a components link's are determined by where its being used, where its really handy to be able to just pass it a resolved href from the layout that is rendering it. 

### What actually happened?

I'm not able to do this which makes this rule very tedious to use

### Link to **GitHub Repo** with Minimal Reproducible Example

Repro not needed as its shown in simple code above

marekdedic:
Thanks, that seems like a false positive, will take a closer look :)

polyzen:
Also perhaps worth noting, this rule does not seem to trigger if you use `<a {href}>foo</>`.

Edit: does **not***

marekdedic:
@polyzen aah, you think you thought of all the edge cases and then after it makes it to recommended, there's so many more.... :D

In other words, thanks, I'll take a look.

cdcarson:
This should be an opt-in rule. It's a breaking change for ci tasks that require linting to pass.

marekdedic:
Every change to the recommended rule set is a breaking change for CI tasks and as far as I understand, the policy of this package is that that is okay in minor versions.

cdcarson:
> Every change to the recommended rule set is a breaking change for CI tasks

True. Every change is **potentially** breaking to somebody's project.  But not every change is created equal:

1. Scope. This change affects _every SvelteKit project_ with `<a href="/foo">Foo</a>` links. It therefore has a very large blast zone.
[...]

PR #1323, opened 2025-09-04
feat(no-navigation-without-resolve): checking link shorthand attributes

marekdedic:
Fixes the issue mentioned in https://github.com/sveltejs/eslint-plugin-svelte/issues/1319#issuecomment-3253049286

But this is too interconnected with #1322, @ota-meshi can we get that one merged first please and I will then rebase this one :)

### DIFF
--- a/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
+++ b/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
@@ -49,6 +49,12 @@ export default createRule('no-navigation-without-resolve', {
 	},
 	create(context) {
 		let resolveReferences: Set<TSESTree.Identifier> = new Set<TSESTree.Identifier>();
+
+		const ignoreGoto = context.options[0]?.ignoreGoto ?? false;
+		const ignorePushState = context.options[0]?.ignorePushState ?? false;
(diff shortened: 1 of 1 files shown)


marekdedic:
This one is now ready :)

marekdedic:
Those are already checked without this PR :) This one just adds support for the shorthand...

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -84,6 +84,26 @@ export default createRule('no-navigation-without-resolve', {
 					}
 				}
 			},
+			SvelteShorthandAttribute(node) {
+				if (
+					context.options[0]?.ignoreLinks === true ||
```
ota-meshi, line 89: Could you extract this outside of the `return {...}` and store it in a variable?
This will prevent us from having to modify the default option in multiple places in the future and forgetting to do so.

ota-meshi (changes requested):

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
marekdedic: Done :)

ota-meshi (approved):
LGTM! Thank you!

PR #1326, opened 2025-09-04
feat(prefer-svelte-reactivity): reporting returned variables

marekdedic:


baseballyama (commented):
Why should we be forced to use `SvelteSet` or `SvelteMap` even when the values aren’t changing?

marekdedic:
Firstly, this PR is part of a wider effort to move the rule to better detect when variables are "encapsulated" inside a function/class and not report them - this PR is an off-shoot of #1287 

But this should make sense even on its own anyway - but I think it does. Example

`fileA`:
```ts
function makeSet() {
  const a = new Set();
  // Populate the set etc.
  return a;
}

export makeSet;
```

`fileB`:
```svelte
<script>
  import { makeSet } from "fileA";

  const a = makeSet();
</script>

<!-- Print the set here -->

<button onclick={a.add(42);}>Click me!</button>
```

In this example, this PR reports the incorrect usage of `Set`. I hope to pair it with the rest of #1287 to ignore all encapsulated uses by default so that we don't get too many FPs.

packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-svelte-reactivity/invalid/arrow-function-return01-input.svelte
```diff
@@ -0,0 +1,8 @@
+<script>
+	const fn = () => {
+		const variable = new Set([1, 2, 1, 3, 3]);
+		return variable;
+	}
+</script>
+
+{fn().has(42)}
```
ota-meshi, line 8: Hmm. This is actually fine as is, right? (It shouldn't be reported, right?)
Could you change it to track the value the function returns, check if there's a mutation, and only report that?

marekdedic: I've been thinking about this more and you are right. I think this needs a change of approach. I will probably make an alternative PR.

PR #1339, opened 2025-09-09
refactor(release): use node24, remove NPM_TOKEN 

dominikg:
and align permissions setup with other svelte org release.yml to enable oidc publishing

### DIFF
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -5,13 +5,14 @@ on:
     branches:
       - main
 
-permissions:
-  contents: write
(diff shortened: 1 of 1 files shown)


dominikg:
fyi @ota-meshi  we are moving all svelte org packages to use oidc publishing to avoid using long lived `NPM_TOKEN` secrets. 

This setup requires configuration of the package on npmjs.com. @Rich-Harris   is going to take care of that

dominikg:
draft until npm package settings are updated

PR #1341, opened 2025-09-10
chore: add `workflow_dispatch` for release action

baseballyama:


baseballyama:
I realized this setting wasn’t necessary since the workflow is triggered every time we push to main.
There might be rare cases where I want to retry after changing the npm settings, but for now it shouldn’t be needed.

PR #1344, opened 2025-09-10
PoC: preventing infinite loops idea

ota-meshi:


baseballyama:
This is exactly the kind of implementation I had in mind for managing visited nodes.

marekdedic:
Thanks for the quick PR. My thoughts comparing it to #1344:
- This would be easier to understand than the `ASTSearchHelper`
- If I understand it correctly, the new function would track all nodes visited from the calling function (hence why it's the first parameter)
    - Do we need to clean up the set in order for it now to grow enormously on large codebases?
    - What if we call `findVariable` from the same function multiple times?
- The `findVariable` function is already re-implemented in this package, so maybe we wouldn't even need `findVaraibleSafe` and could just change `findVariable`? Then we also wouldn't need the internal rule.
- In the previous PR, I was going for a more general pattern inspired by the `ReferenceTracker` from `eslint-utils`. I still think something like that would make sense, but I agree that it would be better to separate it from the `findVariable` recursion issue.
- Is there any other way we could get an infinite loop? The previous PR was in that sense more general and would prevent recursion through any means. But I can't think of another way right now.

ota-meshi:
At least nobody is opposed to this change, so I'll mark it as ready.

> - The `findVariable` function is already re-implemented in this package, so maybe we wouldn't even need `findVaraibleSafe` and could just change `findVariable`? Then we also wouldn't need the internal rule.

Yeah, I think it's worth considering. In this PR, I wanted to minimize the difference and check that it can be checked with an internal custom rule.

> - In the previous PR, I was going for a more general pattern inspired by the `ReferenceTracker` from `eslint-utils`. I still think something like that would make sense, but I agree that it would be better to separate it from the `findVariable` recursion issue.

In my experience, I don't think the internal implementation of ReferenceTracker is very common in normal rule implementations. In other words, I currently think that maintaining the current implementation method as much as possible will make it easier for people with experience with eslint-plugins other than eslint-plugin-svelte to follow.

>  Is there any other way we could get an infinite loop? The previous PR was in that sense more general and would prevent recursion through any means. But I can't think of another way right now.

In my experience, I've never seen an infinite recursion issue like this outside of identifier tracking.

---

This PR owes a lot to your PR #1338, which gave me a lot of inspiration. Thank you so much!

ota-meshi:
When I was testing this, I found that the second lookup to the same identifier node (which is not recursive) returned null, which didn't work :sweat: 
I'll think of a way to solve the problem.

ota-meshi:
I've done a major rewrite to solve the problem.
It's now handled in a pretty basic way.
I think this is easier than having to pass around a set of identifiers.

marekdedic:
Hi @ota-meshi , I just found out the internal rule doesn't trigger here, is that correct? 

https://github.com/sveltejs/eslint-plugin-svelte/blob/main/packages%2Feslint-plugin-svelte%2Fsrc%2Frules%2Fno-navigation-without-resolve.ts#L134

ota-meshi:
Yeah, if there is no recursion it will not be reported.

PR #1372, opened 2025-09-23
Explain the no-navigation-without-resolve rationale

nickolay:
I was completely baffled by the explanation at https://sveltejs.github.io/eslint-plugin-svelte/rules/no-navigation-without-resolve/ , since I had no idea there was even a possibility to configure the base URL of an app.

Before finding the docs I assumed the "Found a link with a url that isn't resolved." message shown in VS code meant it couldn't find my API route (cf. https://github.com/sveltejs/kit/issues/13195), which I was accessing like this: `<a href="/report/download>Download</a>` — perhaps "Found a link with a URL that would break if the app is configured with a non-root base path" would be clearer?

### DIFF
--- a/docs/rules/no-navigation-without-resolve.md
+++ b/docs/rules/no-navigation-without-resolve.md
@@ -14,7 +14,7 @@ since: 'v3.12.0'
 
 ## :book: Rule Details
 
-This rule reports navigation using HTML `<a>` tags, SvelteKit's `goto()`, `pushState()` and `replaceState()` functions without resolving a relative URL. All four of these may be used for navigation, with `goto()`, `pushState()` and `replaceState()` being intended solely for internal navigation (i.e. not leaving the site), while `<a>` tags may be used for both internal and external navigation. When using any way of internal navigation, the URL must be resolved using SvelteKit's `resolve()`, otherwise the site may break. For programmatic navigation to external URLs, using `window.location` is advised.
+This rule reports navigation using HTML `<a>` tags, SvelteKit's `goto()`, `pushState()` and `replaceState()` functions without resolving a relative URL. All four of these may be used for navigation, with `goto()`, `pushState()` and `replaceState()` being intended solely for internal navigation (i.e. not leaving the site), while `<a>` tags may be used for both internal and external navigation. When using any way of internal navigation, the URL must be resolved using SvelteKit's `resolve()`, otherwise the site may break when [configured to live on a non-root path](https://svelte.dev/docs/kit/configuration#paths). For programmatic navigation to external URLs, using `window.location` is advised.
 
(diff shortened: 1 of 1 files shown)


baseballyama:
@nickolay How about https://github.com/sveltejs/eslint-plugin-svelte/pull/1377? I didn’t change the detailed rule description, but I clarified the error messages themselves.

marekdedic:
@baseballyama Why not both? :D I actually like both :)

baseballyama:
I think the correct approach to the original issue is to improve the rule so that it doesn’t report an error when either `data-sveltekit-reload` or `rel="external"` is present.
It’s fine to merge this PR if it’s useful, but at least for me, I still don’t understand the meaning of the `configured to live on a non-root path` properly, so I’d appreciate it if someone could provide additional clarification.

nickolay:
@baseballyama I like your change to the linter messages, as it removes the source of my original confusion. I see my doc change to be complementary, as I’m trying to clarify why the rule exists at all.

The added [link](https://svelte.dev/docs/kit/configuration#paths) points to paths.base documentation, which I believe adequately explains the need for resolve(): 

> [paths.base] specifies where your app is served from and allows the app to live on a non-root path. Note that you need to prepend all your root-relative links with the base value or they will point to the root of your domain, not your base

In other words, if multiple apps coexist on a single domain a page belonging to app1 https://my.company/app1/page1 must prepend `paths.base` (`== “/app1”`) to any internal links. Failing to do that (`<a href=“/page2”>`) will point to https://my.company/page2 instead of /app1/page2.

I didn’t think the rule docs is the right place to go into such details, but if you meant the wording I chose is unhelpful in getting the point across, perhaps this is better:

> When using any way of internal navigation, the URL must be resolved using SvelteKit's `resolve()`, otherwise the site may break **when [paths.base](https://svelte.dev/docs/kit/configuration#paths) is changed to be non-empty**.

I don’t care much about the wording or this PR, so please feel free to just make the changes you want, if that’s easier.

nickolay:
> I think the correct approach to the original issue is to improve the rule so that it doesn’t report an error when either data-sveltekit-reload or rel="external" is present.

If original issue refers to my API endpoint problem (https://github.com/sveltejs/kit/issues/13195), I believe I’d actually need the `resolve()` if I ever decide to use non-empty `paths.base`.

(On the other hand, I would argue that supporting non-empty `base` should be opt-in — e.g. if you have actually configured such a base, — and not on by default.)

sacrosanctic:
It's not just about base path, it also type checks to ensure the internal path is valid. That is arguably the more important feature.

sacrosanctic:
closed by #1500

ota-meshi:
This PR was probably resolved in #1479. I'm closing this PR. Thank you!

Issue #1242, opened 2025-06-02
Rule valid-compile doesn't report when runes: true


uh-zuh:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.

### What version of ESLint are you using?

9.28.0

### What version of `eslint-plugin-svelte` are you using?

3.9.1

### What did you do?

<details>
<summary>Configuration</summary>

```
import js from "@eslint/js";
import sveltePlugin from "eslint-plugin-svelte";
import svelteParser from "svelte-eslint-parser";
import tseslint from "typescript-eslint";
import svelteConfig from './svelte.config.js';

export default tseslint.config(
    js.configs.recommended,
    ...tseslint.configs.recommended,
    ...sveltePlugin.configs["flat/recommended"],
    {
        files: ["**/*.svelte"],
        languageOptions: {
            parser: svelteParser,
            parserOptions: {
                parser: {
                    ts: "@typescript-eslint/parser",
                },
                extraFileExtensions: [".svelte"],
                svelteConfig,
            },
        },
        rules: {
            'svelte/valid-compile': [
                'error',
                {
                    ignoreWarnings: false
                }
            ],
        },
    },
);
```
</details>

After forcing runes mode with `runes: true` in svelte.config.js, vite catchs new warnings but ESLint with `svelte/valid-compile` rule doesn't catch new warnings.

From there https://github.com/uh-zuh/eslint-plugin-svelte_bug_reproduction_valid-compile
I run:
```
npm run build
```
and see 2 errors:
```
npm run build

> build
> vite build --config vite.config.ts

vite v6.3.5 building for production...
transforming (6) node_modules/svelte/src/internal/shared/utils.js10:14:26 PM [vite-plugin-svelte] src/Test1.svelte:3:8 `count` is updated, but is not declared 
with `$state(...)`. Changing its value will not correctly trigger updates
https://svelte.dev/e/non_reactive_update
1: <script lang="ts">
2:
3:     let count = 0;
                ^
4:     function increment() {
5:         count += 1;
10:14:27 PM [vite-plugin-svelte] src/Test2.svelte:3:8 `count` is updated, but is not declared with `$state(...)`. Changing its value will not correctly trigger updates
https://svelte.dev/e/non_reactive_update
1: <script lang="ts">
2:     let { param } = $props();
3:     let count = 0;
                ^
[...]

PR #1373, opened 2025-09-23
fix(valid-compile): use compiler options if provided

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1242


packages/eslint-plugin-svelte/src/shared/svelte-compile-warns/index.ts
```diff
@@ -407,6 +407,7 @@ function getWarningsFromCode(
 } {
 	try {
 		const result = compiler.compile(code, {
+			...context.sourceCode.parserServices.svelteParseContext?.svelteConfig?.compilerOptions,
```
ota-meshi, line 410: To avoid unintended behavior, can you please pass only the necessary options?

Since `generate: false` is different from the compiler used by the user, I think it is highly likely that the user's defined `compilerOptions` cannot be used as is.

ota-meshi (changes requested):

packages/eslint-plugin-svelte/src/shared/svelte-compile-warns/index.ts
ota-meshi: Compiler options have many properties, and I don't think we understand how they affect linting. Also, I don't think users set compiler options with linting in mind.
Therefore, I think it's better to avoid passing them directly.
I think it's better to pass only the necessary properties that we understand.
https://svelte.dev/docs/svelte/svelte-compiler#CompileOptions
https://svelte.dev/docs/svelte/svelte-compiler#ModuleCompileOptions

baseballyama: > I don't think we understand how they affect linting

I believe I understand your concern. I have no objection to making the fix.
However, in this particular case, do we really need to worry about the affect? Wouldn’t it be fine to simply compile exactly as the user specified and then display whatever warnings are generated to the user? If something strange happens, wouldn’t that just mean the user’s configuration was wrong?

ota-meshi: The reason I'm concerned may be because I don't know the internals of the compiler, so I don't have the information to not worry 😓

PR #1377, opened 2025-09-23
fix(no-navigation-without-resolve): improve error messages

baseballyama:
Alternative of https://github.com/sveltejs/eslint-plugin-svelte/pull/1372


baseballyama:
I’ll merge this ahead of time since I want to make some fixes related to this rule.

Issue #1369, opened 2025-09-22
False positive svelte/no-unused-props when spreading props since 3.12.4


machadinhos:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.

### What version of ESLint are you using?

9.36.0

### What version of `eslint-plugin-svelte` are you using?

3.12.4

### What did you do?

<details>
<summary>Configuration</summary>

The default configuration after creating a svelte app with `npx sv create project_name` and choosing typescript and eslint.
</details>

```svelte
<!--Test.svelte-->
<script lang="ts">
	interface Props {a: string;}

	let { a, b }: Props = $props();
</script>

{a}
{b}
```

```svelte
<!--Test1.svelte-->
<script lang="ts">
	import Test from '$lib/Test.svelte';

	interface Props {a: string;}

	let props: Props = $props();
</script>

<Test {...props} />
```

On the `Test1.svelte` component props will have the following error:
```
'a' is an unused Props property  svelte/no-unused-props
```

Changing to `eslint-plugin-svelte@3.12.3` makes the error go away.

Thank you!

### What did you expect to happen?

It shouldn't give an error because `"a"` is in `props` and props is being spread to `<Test1>`. 

### What actually happened?

```
'a' is an unused Props property  svelte/no-unused-props
```

### Link to **GitHub Repo** with Minimal Reproducible Example

https://github.com/machadinhos/no-unused-props-bug-demo

baseballyama:
@machadinhos Are you using Windows?

machadinhos:
@baseballyama yes I am, does that make a difference? 
I know windows is horrible, but don't tell me it's that bad hahaha

baseballyama:
After reviewing the code, I found it’s a bug that also occurs on macOS. It looks like no one had hit this case before.
I’m working on it now👍

Since the GitHub belongs to Microsoft, I can’t really badmouth Windows here😆

PR #1378, opened 2025-09-23
fix(no-unused-props): validate spread operator properly

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1369

I’m not entirely satisfied with this fix, but making the code better would take quite a bit more time, so I’m submitting the PR as-is for now.


ota-meshi (approved):
Thank you!

Issue #1353, opened 2025-09-11
Less strict `no-navigation-without-resolve` default


jycouet:
### Description

Since `no-navigation-without-resolve` is now in recommended (thx to https://github.com/sveltejs/eslint-plugin-svelte/pull/1308) it's triggering some discussions in different places.

In `sv` -> https://github.com/sveltejs/cli/issues/702
About tutorial -> https://github.com/sveltejs/eslint-plugin-svelte/pull/1308#issuecomment-3279043029
Here 1 -> https://github.com/sveltejs/eslint-plugin-svelte/issues/1327
Here 2 -> https://github.com/sveltejs/eslint-plugin-svelte/issues/1319

Maybe, at the moment, the default config is a bit too aggressive.
What do you think about relaxing it a bit like:
```diff
{
  "svelte/no-navigation-without-resolve": [
    "error",
    {
      "ignoreGoto": false,
-      "ignoreLinks": false,
+      "ignoreLinks": true,
      "ignorePushState": false,
      "ignoreReplaceState": false
    }
  ]
}
```

Let me know

marekdedic:
I think we should fix the issues with the rule (working on it, but PRs/reviews are slow-ish :( ) and then decide. Also, I don't see why links should be ignored and the others left turned on.

jycouet:
It looks like quite a few issues have been raised recently regarding `no-navigation-without-resolve`, probably because it’s now part of the recommended set.

I was thinking it might be better to make the rule a bit less aggressive while it’s still fresh. That way, adoption could be smoother in the long run. (as `ignoreLinks` is the first hit, I was suggesting this one, but maybe all reverted for now could be an option too)

Relaxing the recommendation a little would also mean we don’t have to change defaults later on (in ou projects), and it would give us _(You? ^^, maybe me? :) )_ time to improve things in a more relaxed manner.

joeychilson:
> I think we should fix the issues with the rule (working on it, but PRs/reviews are slow-ish :( ) and then decide. Also, I don't see why links should be ignored and the others left turned on.

I think the problem people are running into with links compared to the other options is links can be external and internal and I don't think it possible to resolve a link that's optional/undefined because there is not way to know if the link is external or internal. I think this is the reason someone would ignore links.

hyunbinseo:
I am having problems with external links as well.

Sure, I can ignore them individually, but it's a pain.

```svelte
<script>
[...]

PR #1380, opened 2025-09-23
fix(no-navigation-without-resolve): do not report if there is `data-sveltekit-reload` or `rel="external"` in `<a>`

baseballyama:
close part of https://github.com/sveltejs/eslint-plugin-svelte/issues/1353

There haven’t been any reports yet, but I realized we’re missing handling for another way of specifying external links.

### DIFF
--- a/.changeset/brown-lights-show.md
+++ b/.changeset/brown-lights-show.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': patch
+---
+
+fix(no-navigation-without-resolve): do not report if there is `data-sveltekit-reload` or `rel="external"` in `<a>`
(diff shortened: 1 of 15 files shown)


marekdedic:
I understand `rel="external"` (and it was actually reported in some related issue), but not the `data-sveltekit-reload` - why should that one be treated as external?

sacrosanctic:
`data-sveltekit-reload` causes a hard refresh, but if the target is an internal link, it should still be linted.

marekdedic:
@baseballyama Most of this PR (the `rel="external"` part) has been solved in the meantime by #1443. As @sacrosanctic points out, `data-sveltekit-reload` shouldn't be automatically allowed.

So I think this can be closed, right? Thank you for it nonetheless, I used parts of this in #1433 :smile:

Issue #1327, opened 2025-09-05
no-navigation-without-resolve doesn't work with search params


sillvva:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.

### What version of ESLint are you using?

9.34.0

### What version of `eslint-plugin-svelte` are you using?

3.12.1

### What did you do?

<details>
<summary>Configuration</summary>
// ESLint flat config for SvelteKit + TypeScript + Prettier
// Migrated from .eslintrc.cjs to eslint.config.js (Flat Config)

import js from "@eslint/js";
import eslintConfigPrettier from "eslint-config-prettier/flat";
import svelte from "eslint-plugin-svelte";
import globals from "globals";
import svelteConfig from "./svelte.config.js";

// Use the typescript-eslint aggregator for flat config presets
// (requires devDependency: "typescript-eslint")
import tseslint from "typescript-eslint";

export default [
	// Ignore generated and build artifacts
	{
		ignores: [
			// migrated from .eslintignore
			".DS_Store",
			"node_modules",
			"build",
			".svelte-kit",
			"package",
			".env",
			".env.*",
			"pnpm-lock.yaml",
			"package-lock.json",
			"yarn.lock",
			// additional
			"dist",
			".vite",
			"drizzle",
			"supabase/**/migrations/**"
		]
	},

	// Base language options and globals
	{
		languageOptions: {
			ecmaVersion: 2020,
			sourceType: "module",
			globals: {
				...globals.browser,
				...globals.node,
				// Svelte built-in types
				$$Generic: "readonly",
				App: "readonly"
			}
		}
	},

	// Core JS recommended rules
	js.configs.recommended,

	// TypeScript rules (no type-aware rules by default)
	...tseslint.configs.recommended,

	// Svelte plugin recommended rules for Svelte files
	...svelte.configs["flat/recommended"],
	{
		files: ["**/*.svelte", "**/*.svelte.ts", "**/*.svelte.js"],
		languageOptions: {
			parserOptions: {
				// Ensure TS within <script lang="ts"> is parsed
				parser: tseslint.parser,
				projectService: true,
				extraFileExtensions: [".svelte"],
				// Pass through SvelteKit config for better rule behavior
				svelteConfig
			}
		}
	},
	// Turn off formatting-related Svelte rules to avoid conflicts with Prettier
[...]

PR #1381, opened 2025-09-24
feat(no-navigation-without-resolve): add `allowSuffix` option

baseballyama:
close part of https://github.com/sveltejs/eslint-plugin-svelte/issues/1353
close part of https://github.com/sveltejs/eslint-plugin-svelte/issues/1327

### DIFF
--- a/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
+++ b/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
@@ -1,5 +1,6 @@
 import type { TSESTree } from '@typescript-eslint/types';
 import { createRule } from '../utils/index.js';
+import type { TrackedReferences } from '@eslint-community/eslint-utils';
 import { ReferenceTracker } from '@eslint-community/eslint-utils';
 import { FindVariableContext } from '../utils/ast-utils.js';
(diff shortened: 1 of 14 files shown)


marekdedic:
I think this should be the default behaviour. I'll take a look at the code in a few days - I have some other ideas for improvements as well, so would like to push it in a certain way...

marekdedic:
Hi,
thanks for the PR, it generally looks like the right idea. Several points about the design

1. I think this doesn't need an option gating it, suffixes should always be allowed
2. I think this should be checked for both `resolve()` as well as `asset()` - while not common, even assets can have query options.

Those 2 should simplify the code significantly. On the other hand, one more complication:

3. I don't think we should just allow any suffix. My first idea would be to check that the suffix starts with either a `?` or a `#` - I think that would be a reasonable trade-off between complexity and correctness

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
marekdedic, line 381: I think this PR should remove/repurpose this function

marekdedic, line 434: I think this PR should remove/repurpose this function

marekdedic (changes requested):

PR #1383, opened 2025-09-26
fix(no-navigation-without-resolve): do not report parent-directory relative urls in anchor links

yanneves:
updated (without refactor): https://github.com/sveltejs/eslint-plugin-svelte/pull/1383#issuecomment-4020166705

### DIFF
--- a/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
+++ b/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
@@ -227,6 +227,7 @@ function checkLinkAttribute(
 		!isValueAllowed(new FindVariableContext(context), value, resolveReferences, {
 			allowAbsolute: true,
 			allowFragment: true,
+			allowParentDirectoryRelative: true,
 			allowNullish: true
 		})
 	) {
@@ -277,6 +278,7 @@ function isValueAllowed(
 		allowAbsolute?: boolean;
 		allowEmpty?: boolean;
 		allowFragment?: boolean;
+		allowParentDirectoryRelative?: boolean;
 		allowNullish?: boolean;
 	}
 ): boolean {
@@ -295,6 +297,7 @@ function isValueAllowed(
 		(config.allowAbsolute && expressionIsAbsoluteUrl(ctx, value)) ||
 		(config.allowEmpty && expressionIsEmpty(value)) ||
 		(config.allowFragment && expressionStartsWith(ctx, value, '#')) ||
+		(config.allowParentDirectoryRelative && expressionStartsWith(ctx, value, '..')) ||
 		(config.allowNullish && expressionIsNullish(value)) ||
 		expressionIsResolveCall(ctx, value, resolveReferences)
 	) {
--- a/packages/eslint-plugin-svelte/tests/fixtures/rules/no-navigation-without-resolve/valid/link-relative-url01-input.svelte
+++ b/packages/eslint-plugin-svelte/tests/fixtures/rules/no-navigation-without-resolve/valid/link-relative-url01-input.svelte
@@ -1,0 +1,11 @@
+<script>
+	const value = '../';
+	const href = '../';
+</script>
+
+<a href="..">Click me!</a>
+<a href="../">Click me!</a>
+<a href="../other">Click me!</a>
+<a href={'..' + '/other'}>Click me!</a>
+<a href={value}>Click me!</a>
+<a {href}>Click me!</a>


packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -263,37 +261,37 @@ function expressionIsEmpty(url: TSESTree.CallExpressionArgument): boolean {
...
-function binaryExpressionIsAbsolute(
+function binaryExpressionIsIgnored(
 	ctx: FindVariableContext,
 	url: TSESTree.BinaryExpression
 ): boolean {
 	return (
-		(url.left.type !== 'PrivateIdentifier' && expressionIsAbsolute(ctx, url.left)) ||
-		expressionIsAbsolute(ctx, url.right)
+		(url.left.type !== 'PrivateIdentifier' && expressionIsIgnored(ctx, url.left)) ||
+		expressionIsIgnored(ctx, url.right)
```
marekdedic, line 290: Beware that this would change the behaviour for the expression `"a" + "#b"` which currently isn't ignored but would be in this PR

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -303,73 +301,32 @@ function identifierIsAbsolute(ctx: FindVariableContext, url: TSESTree.Identifier
...
 
-function templateLiteralIsAbsolute(
+function templateLiteralIsIgnored(
 	ctx: FindVariableContext,
 	url: TSESTree.TemplateLiteral
 ): boolean {
 	return (
-		url.expressions.some((expression) => expressionIsAbsolute(ctx, expression)) ||
-		url.quasis.some((quasi) => urlValueIsAbsolute(quasi.value.raw))
+		url.expressions.some((expression) => expressionIsIgnored(ctx, expression)) ||
```
marekdedic, line 313: Similarly to binary expression, this changes the behaviour for fragment links.

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -303,73 +301,32 @@ function identifierIsAbsolute(ctx: FindVariableContext, url: TSESTree.Identifier
...
-	);
+function urlValueIsAbsolute(url: string): boolean {
+	return /^[+a-z]*:/i.test(url);
 }
 
 function urlValueIsFragment(url: string): boolean {
 	return url.startsWith('#');
 }
+
+function urlValueIsUpperDirectory(url: string): boolean {
```
marekdedic, line 330: Can you describe why this should be ignored?

marekdedic (changes requested):
Hi,
I think this PR subtly changes some behaviours where I'd like to make sure we are doing that deliberately and not by mistake. I can see 2 ways in which this PR would change behaviours:

1. Currently fragment links have to _start_ with `#`. In this PR, this is relaxed to some _part_ of the expression having to start with `#`. This does not make sense to me. There is also the discussion on #1327 (which proposes a reasonable change to the rule), but this isn't that.
2. There is an added exception for links where any part of the expression starts with `..` (e. g. `"a" + "..b"`). I don't really see why these should be ignored, but maybe I'm missing something...

To address the original rationale: I agree it would be better to only traverse once, but when I was implementing this I didn't find a way to do it while preserving the current behaviour and not making it spaghetti code.

marekdedic:
Hi, thank you for the PR, however, I think this has mostly been superseded by #1476 and can be closed, right?

yanneves:
nice one, appreciate the big refactor @marekdedic it's better than my attempt + sorry for my bad etiquette, I got carried away with the refactor originally when trying to introduce this change, didn't provide background as to why parent-directory relative urls present an issue, and took 6 months to circle back 🫣

I've just rebased this branch, removes any refactor and now just adds an allowed case for [parent-directory relative urls](https://developer.mozilla.org/en-US/docs/Web/API/URL_API/Resolving_relative_references#parent-directory_relative) (i.e. `<a href="../">` along with test fixtures

without this change we report these urls to be unexpected without `resolve()`

attempting to wrap `<a href="../">` as `<a href={resolve("../")}>` changes the output from `"../"` to `"/./"` and breaks the link

without using `resolve()` these urls pass through and are also unaffected in SSR - they might be ignored intentionally

an example of where you might use this url structure: say you have a blog on `example.com/blog/` and posts are on `example.com/blog/first-post/` etc, then to decouple the layout for posts from your routes you can use `<a href="../">` on a post instead of `<a href="/blog/">` to go back up to view all posts, which avoids having to find-and-replace all root-level links if you moved the route from `/blog` to `/articles`

a simple example but the problem compounds in more complex route hierarchies

marekdedic:
Thanks for the rebase.

I am not sure that your proposal is sound in that a relative navigation (`href="../"`) is not guaranteed to be internal and valid. It *may* be (such as in your example), but it may also not be (depending on the depth of the route and the href). In your example, I don't really see a big benefit of using relative paths (it looks neater, but is more fragile).

Also, I don't want to beat you with process too much, but I think next time creating an issue first (and having this discussion making it more precise) and only then opening a PR is the preferred way.

Issue #1434, opened 2025-11-29
cannot use await false warning even though `experimental.async` is turned on


that-richan:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.

### What version of ESLint are you using?

9.39.1

### What version of `eslint-plugin-svelte` are you using?

3.13.0

### What did you do?

<details>
<summary>Configuration</summary>

svelte.config.js
```
import { vitePreprocess } from "@sveltejs/vite-plugin-svelte";

export default {
  compilerOptions: {
    experimental: {
      async: true,
    },
  },
  preprocess: [vitePreprocess()],
};

```

eslint.config.js
```
import { defineConfig } from "eslint/config";
import js from "@eslint/js";
import svelte from "eslint-plugin-svelte";
import globals from "globals";
import ts from "typescript-eslint";
import svelteConfig from "./svelte.config.js";

export default defineConfig([
  js.configs.recommended,
  ...ts.configs.recommended,
  ...svelte.configs.recommended,
  {
    languageOptions: {
      globals: {
        ...globals.browser,
        ...globals.node,
      },
    },
  },
  {
    files: ["**/*.svelte", "**/*.svelte.ts", "**/*.svelte.js"],
    languageOptions: {
      parserOptions: {
        projectService: true,
        extraFileExtensions: [".svelte"],
        parser: ts.parser,
        svelteConfig,
      },
    },
    rules: {
      "svelte/valid-compile": "error",
      "svelte/no-target-blank": "error",
      "svelte/block-lang": [
        "error",
        {
          enforceScriptPresent: true,
          enforceStylePresent: false,
          script: ["ts"],
          style: [null],
        },
      ],
      "svelte/button-has-type": "error",
      "svelte/no-add-event-listener": "error",
      "svelte/no-ignored-unsubscribe": "error",
      "prefer-const": "off",
      "svelte/prefer-const": [
        "error",
        {
          destructuring: "all",
          excludedRunes: [],
        },
      ],
      "svelte/prefer-destructured-store-props": "error",
      "svelte/require-stores-init": "error",
      "svelte/no-top-level-browser-globals": "error",
    },
  },
]);

```
</details>

```svelte
<script lang="ts">
	/*
	* Cannot use `await` in deriveds and template expressions, or at the top level of a component, unless the `experimental.async` compiler option is `true`
[...]

PR #1438, opened 2025-12-04
fix: Support `experimental.async` compiler option

baseballyama:
close: https://github.com/sveltejs/eslint-plugin-svelte/issues/1434


ota-meshi (approved):
Thanks!

PR #1460, opened 2026-01-14
docs(no-navigation-without-resolve): add <a rel="external" ...> example

langpavel:
This rule behavior is not obvious from description and is implemented.

There is also https://github.com/sveltejs/eslint-plugin-svelte/pull/1380 with more work done, but with conflicts.


ota-meshi (approved):
Thank you!

PR #1467, opened 2026-01-23
fix(no-navigation-without-resolve): fixed nullish link shorthands not being allowed

marekdedic:
Small one, but needed to be able to unify the logic in the great refactoring of this rule...


marekdedic:
Thanks! No worries, but I'm afraid the next one is rather bigger: #1476

PR #1444, opened 2025-12-14
Reworked no-navigation-without-resolve to be more general and enable more edge cases

marekdedic:
Fixes #1327

PR #1482, opened 2026-03-04
feat(no-navigation-without-resolve): recognizing resolved values with queries and fragments

### LINKED ISSUE #1327: no-navigation-without-resolve doesn't work with search params

sacrosanctic:
@marekdedic

a refactor of #1444
closes #1327

I moved `hasRelExternal` to reduce the git diff count.


marekdedic:
**THIS SHOULD NOT BE MERGED YET**

Hi, the code looks good (it's mostly mine, heh :sweat_smile:), so I'm OK with the code (could be split into 2 PRs, but whatever, they are in 2 commits...). However, I just found that the Svelte team is considering adding this functionality into `resolve()` (and there is as of now merged but unreleased PR that adds support for passing query and fragment as strings into `resolve()`). I have asked them how they want to proceed and I'd like to wait fro their answer. Maybe we shouldn't do this PR after all as the functionality will be included in `resolve()`...

sacrosanctic:
I saw that, what an unfortunate timing on my part. I just wanted to push progress forward.

marekdedic:
@sacrosanctic Can you add a "Closes #1327"?

jestersoftware:
I'm confused on why this hasn't moved I must be missing something @marekdedic 

Right now I cannot do this very basic thing:

`<a href={'/abc?x=' + '1'}` (or equivalent where the '1' is a variable)
-> it complains about not using resolve

add resolve
`<a href={resolve('/abc?x=' + '1')}`
-> resolve works fine and provides the correct link but it gives `ts-2345` error

Only solution afaik is 1) not using `resolve` (bad) and therefore then 2) adding `eslint-disable-next-line svelte/no-navigation-without-resolve` so things don't light up "red" in IDE. This is what I'm forced to do at present to do a simple `a` with a query param.

So this definitely seems like a linting issue, and has nothing to do with `resolve()` and has nothing to do with `goto()`: `resolve` works as is and renders a correct "href" on the `a`, as noted above, and `goto` is a totally different programmatic solution that's not the same as a simple html "link" (although it may need similar fixing which I think was done already). Imo regardless of those other solutions this is simply a linting bug as the above URL is totally acceptable in every way and it's a extremely fundamental part of HTML.

Regarding "user could screw up the query param" noted in https://github.com/sveltejs/kit/issues/14750#issuecomment-3433894692.... Sure. Don't relax the lint rule so much that obvious problems are allowed. But imo mostly irrelevant as that can be solved elsewhere. A rule that dissallows the 99% to worry about the 1% is not a good rule imo.

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -270,7 +282,8 @@ function hasRelExternal(ctx: FindVariableContext, element: AST.SvelteStartTag):
 }
 
 function isValueAllowed(
-	ctx: FindVariableContext,
+	context: RuleContext,
+	findContext: FindVariableContext,
```
jycouet, line 300: Either drop the findContext parameter and always construct from context, or keep it and reuse it everywhere

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -281,33 +294,73 @@ function isValueAllowed(
...
+		].sort((a, b) => a.range[0] - b.range[0]);
+		if (
+			expressionIsResolveCall(new FindVariableContext(context), parts[0], resolveReferences) &&
+			(expressionIsEmpty(new FindVariableContext(context), parts[1]) ||
+				expressionStartsWith(new FindVariableContext(context), parts[1], '?') ||
+				expressionStartsWith(new FindVariableContext(context), parts[1], '#'))
+		) {
+			return true;
+		}
+	}
```
jycouet, line 382: Add a guard, e.g. parts.length >= 2 && before the inner check?

jycouet (commented):
Thx for adding all these tests 👍

marekdedic:
This hasn't moved because when I last checked, it seemed that SvelteKit would add support for queries & fragments to `resolve()`. It seems that that is still the plan, but stalled somewhat.

If that is merged, I think we'll want to disallow bypassing resolve with queries and fragments. So I'm a little hesitant to merge this now, as we could well be reverting this change soon-ish. @ota-meshi what do you think?

teemingc:
I don't think we'll be adding those options to `resolve` anytime soon (see https://github.com/sveltejs/kit/issues/14750#issuecomment-3996111011). The current PR that's open doesn't fulfil that requirement. It's probably better to go ahead with this PR as the eslint rule is currently too restrictive but is enabled by default.

marekdedic:
Ok, so I'd be for merging this. LGTM.

marekdedic:
@sacrosanctic Could you rebase please? I think I introduced some merge conflicts, sorry...

sacrosanctic:
@marekdedic I was thinking "I don't remember there being conflicts when I opened this."

Resolved the merge conflicts, I gave `expressionIsResolvedPathname` context, but didn't use it. Unsure if it's needed. These changes does not address @jycouet's comments.

edit: runner didn't like that so I removed it.

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -328,20 +332,54 @@ function isValueAllowed(
...
-			isValueAllowed(ctx, value.alternate, resolveReferences, tsTools, config)
+			isValueAllowed(context, findContext, value.consequent, resolveReferences, tsTools, config) &&
+			isValueAllowed(context, findContext, value.alternate, resolveReferences, tsTools, config)
 		);
 	}
 	if (
-		(config.allowAbsolute && expressionIsAbsoluteUrl(ctx, value)) ||
-		(config.allowEmpty && expressionIsEmpty(value)) ||
-		(config.allowFragment && expressionStartsWith(ctx, value, '#')) ||
+		(config.allowAbsolute && expressionIsAbsoluteUrl(new FindVariableContext(context), value)) ||
```
ota-meshi, line 340: What is the intention behind this change? Could you explain?

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -328,20 +332,54 @@ function isValueAllowed(
...
+		(config.allowEmpty && expressionIsEmpty(new FindVariableContext(context), value)) ||
+		(config.allowFragment && expressionStartsWith(new FindVariableContext(context), value, '#')) ||
 		(config.allowNullish && expressionIsNullish(value)) ||
 		expressionIsResolvedPathname(value, tsTools) ||
-		expressionIsResolveCall(ctx, value, resolveReferences)
+		expressionIsResolveCall(new FindVariableContext(context), value, resolveReferences)
 	) {
 		return true;
 	}
+	if (value.type === 'BinaryExpression' && value.left.type !== 'PrivateIdentifier') {
```
ota-meshi, line 349: Is this intended to handle concat operations using `+`? If so, I think we need to check the operator.

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -328,20 +332,54 @@ function isValueAllowed(
...
+				context,
+				new FindVariableContext(context),
+				value.left,
+				resolveReferences,
+				tsTools,
+				config
+			);
+		}
+		if (
+			expressionIsResolveCall(new FindVariableContext(context), value.left, resolveReferences) &&
```
ota-meshi, line 361: Shouldn't you be using `findContext` instead of `new FindVariableContext()`? Otherwise, I think we'll get an infinite recursion error depending on the code.

ota-meshi (changes requested):
Thank you for this pull request!

I've left a few comments.
Could you also investigate the cause of the error in CI?

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
marekdedic: I also don't really understand this, why don't we just use `ctx`? (then we also wouldn't need to pass `context` around)

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
marekdedic, line 349: I'd move this above the "master if" (L339 currently)

marekdedic: This is probably an old version, I'd recommend following the logic of `binaryExpressionIsAbsoluteUrl`

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -328,20 +332,54 @@ function isValueAllowed(
...
+		(config.allowFragment && expressionStartsWith(new FindVariableContext(context), value, '#')) ||
 		(config.allowNullish && expressionIsNullish(value)) ||
 		expressionIsResolvedPathname(value, tsTools) ||
-		expressionIsResolveCall(ctx, value, resolveReferences)
+		expressionIsResolveCall(new FindVariableContext(context), value, resolveReferences)
 	) {
 		return true;
 	}
+	if (value.type === 'BinaryExpression' && value.left.type !== 'PrivateIdentifier') {
+		if (['BinaryExpression', 'TemplateLiteral'].includes(value.left.type)) {
```
marekdedic, line 350: Hmm, sorry, I already forgot, why this double if?

marekdedic (changes requested):
Sorry, I looked over it again and have some questions/feedback.

Let me know if I should do some work on these as well, don't want to put too much burden on you :/

marekdedic:
@sacrosanctic Sorry to bother, but do you have the time to look at this again? If not, that's okay and I'll do it :)

sacrosanctic:
I'll take a stab at it tonight. If you don't hear from me by then. It's cause I didn't have time.

PR #1491, opened 2026-03-08
docs: slight wording adjustment so it isn't so awkward

Azuriru:
I think the wording is a mistake, if it isn't, please disregard this PR


ota-meshi (approved):
Thank you!

PR #1503, opened 2026-03-31
fix: disallow props based on SvelteKit route component type

teemingc:
This PR:

- removes the unused `errors` prop. This doesn't exist anymore
- discriminates valid props based on file type (+error.svelte accepts the `error` prop but other file types don't. Similarly, +layout.svelte accepts `children` but +page.svelte doesn't)


baseballyama (approved):
Thanks! My understanding is that `errors` was dropped back in the `1.0.0-next` era, so this dates back quite a long way. Given that, I’d classify this as a patch-level change rather than a breaking change.

Issue #1515, opened 2026-04-13
Removed `SourceCode` methods create issues with ESLint 10


kevin-easygo:
### Before You File a Bug Report Please Confirm You Have Done The Following...

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have updated to the latest version of the packages.

### What version of ESLint are you using?

10.2.0

### What version of `eslint-plugin-svelte` are you using?

3.17.0

### What did you do?

<details>
<summary>Configuration</summary>

```ts
// eslint.config.ts

import tseslint from 'typescript-eslint';
import svelte from 'eslint-plugin-svelte';
import svelteParser from 'svelte-eslint-parser';
import { defineConfig } from 'eslint/config';

const config = {
  files: ['**/*.svelte', '**/*.svelte.ts'],
  languageOptions: {
    parser: svelteParser,
    parserOptions: {
      parser: tseslint.parser,
      extraFileExtensions: ['.svelte'],
    },
  },
};

export default defineConfig(
  svelte.configs['flat/recommended'],
  svelte.configs['flat/prettier'],
);
```
</details>

```svelte
<script lang="ts">
  $: test = () => {
    console.log("test");
  };
</script>
```

### What did you expect to happen?

ESLint should catch the `svelte/no-reactive-functions` error and report it.

### What actually happened?

ESLint throws `TypeError: source.isSpaceBetweenTokens is not a function` due to the deprecation of some `SourceCode` methods from ESLint 10 ([reference](https://eslint.org/docs/latest/use/migrate-to-10.0.0#-removal-of-deprecated-sourcecode-methods)).

### Link to **GitHub Repo** with Minimal Reproducible Example

[...]

PR #1516, opened 2026-04-14
fix(eslint-plugin-svelte): Replace `isSpaceBetweenTokens` with `isSpaceBetween`

kevin-easygo:
This PR removes usage of `isSpaceBetweenToken` and replaces it with `isSpaceBetween`, since the former has been deprecated.

Solves #1515.

### DIFF
--- a/packages/eslint-plugin-svelte/src/rules/no-reactive-functions.ts
+++ b/packages/eslint-plugin-svelte/src/rules/no-reactive-functions.ts
@@ -54,7 +54,7 @@ export default createRule('no-reactive-functions', {
 									count: 3
 								});
 
-								const noExtraSpace = source.isSpaceBetweenTokens(
+								const noExtraSpace = source.isSpaceBetween(
 									tokens[1] as AST.Token,
 									tokens[2] as AST.Token
 								);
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-svelte/src/rules/no-reactive-functions.ts
```diff
@@ -54,7 +54,7 @@ export default createRule('no-reactive-functions', {
 									count: 3
 								});
 
-								const noExtraSpace = source.isSpaceBetweenTokens(
+								const noExtraSpace = source.isSpaceBetween(
```
ota-meshi, line 57: Could you please add a test case that confirms this change has resolved the issue?

ota-meshi (changes requested):
Thank you for this PR!

packages/eslint-plugin-svelte/src/rules/no-reactive-functions.ts
kevin-easygo: Hey there! The method has only been deprecated in ESLint 10, while the repo relies on `eslint@~9.39.4`.
What's the best way to test this without changing the dependency list?

ota-meshi: CI runs with ESLint v8/v9/v10 versions.

https://github.com/sveltejs/eslint-plugin-svelte/blob/77d1018d4a30f24a533eb1b3b1c7f7dc36a13cfd/.github/workflows/NodeCI.yml#L61

Therefore, I believe we can verify whether the issue has been resolved simply by adding the code you provided in the issue as a test case.

PR #1518, opened 2026-04-17
feat: add `sort-props` rule

Jungzl:
This PR introduces a new rule, `svelte/sort-props`.
Heavily inspired by [`sort-jsx-props`](https://perfectionist.dev/rules/sort-jsx-props) from `eslint-plugin-perfectionist`, following the same design philosophy


ota-meshi:
I don't want to maintain two rules with the same purpose. Thank you.

PR #1523, opened 2026-05-15
fix(no-navigation-without-resolve): handle optional ResolvedPathname properties

Stadly:
## Problem

The `expressionIsResolvedPathname` function introduced in #1321 uses bidirectional type assignability to check if a value's type matches `ResolvedPathname`:

```ts
checker.isTypeAssignableTo(nodeType, resolvedPathnameType) &&
checker.isTypeAssignableTo(resolvedPathnameType, nodeType)
```

This fails for **optional properties** (e.g. `href?: ResolvedPathname`) because the inferred type is `ResolvedPathname | undefined`, and `ResolvedPathname | undefined` is not assignable to `ResolvedPathname`.

Similarly, accessing a property on an object from an array (e.g. `item.href` where `items: { href?: ResolvedPathname }[]`) produces the same union type.

## Fix

Use `checker.getNonNullableType(nodeType)` to strip `undefined`/`null` from the node type before the assignability comparison:

```ts
const nonNullableNodeType = checker.getNonNullableType(nodeType);
return (
    checker.isTypeAssignableTo(nonNullableNodeType, resolvedPathnameType) &&
    checker.isTypeAssignableTo(resolvedPathnameType, nonNullableNodeType)
);
```

This is safe because the rule already separately handles nullish values via `allowNullish`/`expressionIsNullish` for links. The purpose of `expressionIsResolvedPathname` is to check whether the non-null portion of the type is a valid resolved pathname.

## Tests

Added valid test fixtures covering:
- `href?: ResolvedPathname` prop with shorthand attribute
- `item.href` from an array of objects with optional `href?: ResolvedPathname`
- `goto(href)` with optional `ResolvedPathname` parameter
- `pushState(href, {})` with optional `ResolvedPathname` parameter
- `replaceState(href, {})` with optional `ResolvedPathname` parameter

All of these would produce false positives before this fix.

Closes: related to #1319

### DIFF
--- a/.changeset/fix-optional-resolved-pathname.md
+++ b/.changeset/fix-optional-resolved-pathname.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-svelte": patch
+---
+
+fix(no-navigation-without-resolve): handle optional ResolvedPathname properties in links
--- a/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
+++ b/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
@@ -375,10 +375,13 @@ function expressionIsResolvedPathname(
 	}
 	const resolvedPathnameType = checker.getDeclaredTypeOfSymbol(resolvedPathnameSymbol);
 
+	// Strip nullability so that optional properties (e.g. `href?: ResolvedPathname`) are still recognized.
+	const nonNullableNodeType = checker.getNonNullableType(nodeType);
+
 	// getTypeAtLocation returns the resolved (structural) type without alias information, so we cannot compare aliasSymbols directly. Instead we check structural equivalence by testing assignability in both directions: this correctly rejects strict subtypes like Pathname (Pathname ⊂ ResolvedPathname, so only one direction holds).
 	return (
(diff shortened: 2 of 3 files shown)


marekdedic:
Hi, thanks for the effort to improve the rule.

I'm not sure I understand what you're trying to fix - the current behaviour seems correct to me - `goto()` doesn't take undefined as its argument. 

The best way forward would be, IMO, to open an issue outlining your description of what is incorrect about the rule and go forward from there.

Stadly:
You're right - AI went a bit overboard when writing test cases for me, sorry! I should have noticed.

The issue is specifically about **links**. When a component has `href?: ResolvedPathname`, the type is `ResolvedPathname | undefined`. The rule passes `allowNullish: true` for `<a href>`, but `expressionIsNullish` only catches syntactically nullish expressions, not types that include `undefined`. And `expressionIsResolvedPathname` fails because the bidirectional assignability check doesn't hold for `ResolvedPathname | undefined`.

I've opened #1525 with a detailed description and reproduction. I think the solution in this PR is good, while some of the test cases provided no value.

marekdedic:
@Stadly @ota-meshi I believe this can be closed (superseded by #1532)

Thanks!

PR #1530, opened 2026-05-21
feat: add `no-nested-style-tag` rule

### LINKED ISSUE #1186: new rule: `no-nested-style-tag`
### Motivation

This rule prohibits the use of nested <style> tags to prevent unintended style application.

https://svelte.dev/docs/svelte/nested-style-elements

### Description

Nested <style> tags are not scoped, which may lead to unintended styles being applied.

### Examples

```svelte
<script>
</script>

<!-- ✓ GOOD -->
<style>...</style>

<!-- ✗ BAD -->
<div>
  <style>...</style>
</div>

{@snippet}
  <style>...</style>
{/snippet}
```

### Additional comments

_No response_

SAY-5:
Closes #1186.

Adds a new rule `svelte/no-nested-style-tag` that disallows `<style>` elements nested inside other elements or blocks. Nested `<style>` elements are not scoped by Svelte and may lead to unintended styles being applied, as documented at https://svelte.dev/docs/svelte/nested-style-elements.

Implementation detects any `SvelteElement` whose HTML tag name is `style`. The top-level `<style>` is parsed as a distinct `SvelteStyleElement` by `svelte-eslint-parser` and is therefore not visited, so the rule fires only on the nested case.


ota-meshi (approved):
LGTM! Thank you!

PR #1531, opened 2026-05-21
feat: add `prefer-derived-over-derived-by` rule

### LINKED ISSUE #1246: New rule: `svelte/prefer-derived-over-derived-by`
### Motivation

From a consistency perspective, this rule reports unnecessary usage of `$derived.by`. There is no performance difference. It purely enforces a consistent coding style. Even if included in recommended config, a warning level should be sufficient.

### Description

This rule reports unnecessary usage of `$derived.by`. It also provides a fixer to automatically convert `$derived.by` to `$derived`.

### Examples

```svelte
<script>
// ✓ GOOD
const foo = $derived(a.b);

// ✗ BAD
const bar = $derived.by(() => a.b);
</script>
```

### Additional comments

_No response_

SAY-5:
Closes #1246.

Adds a new rule `svelte/prefer-derived-over-derived-by` that reports unnecessary uses of `$derived.by()` where `$derived()` is sufficient, with an autofix.

The rule flags `$derived.by()` calls whose argument is a zero-parameter, non-async, non-generator function that either:
- uses a concise arrow body (e.g. `(() => a.b)`), or
- uses a block body containing only a single `return` statement (e.g. `(() => { return a.b; })` or `function () { return a.b; }`).

Multi-statement bodies are left untouched because removing the wrapping function would change semantics.


packages/eslint-plugin-svelte/src/rules/prefer-derived-over-derived-by.ts
```diff
@@ -0,0 +1,74 @@
...
+					return;
+				}
+				if (node.arguments.length !== 1) {
+					return;
+				}
+				const arg = node.arguments[0];
+				if (arg.type !== 'ArrowFunctionExpression' && arg.type !== 'FunctionExpression') {
+					return;
+				}
+				if (arg.params.length !== 0 || arg.async || arg.generator) {
```
ota-meshi, line 46: Could you please add some test cases that confirm these are ignored?

packages/eslint-plugin-svelte/src/rules/prefer-derived-over-derived-by.ts
```diff
@@ -0,0 +1,74 @@
...
+				let expressionText: string | null = null;
+				if (arg.type === 'ArrowFunctionExpression' && arg.body.type !== 'BlockStatement') {
+					expressionText = context.sourceCode.getText(arg.body);
+				} else if (
+					arg.body.type === 'BlockStatement' &&
+					arg.body.body.length === 1 &&
+					arg.body.body[0].type === 'ReturnStatement' &&
+					arg.body.body[0].argument !== null
+				) {
+					expressionText = context.sourceCode.getText(arg.body.body[0].argument);
```
ota-meshi: Could you please modify this line and L51 to retain the expression node, and change the text extraction to be done within the `fix()` function?

ESLint sometimes doesn't call the `fix()` function depending on a flag, and I think this would slightly reduce the processing load.

ota-meshi (changes requested):
Thank you for this PR!

ota-meshi (approved):
Wow! Thank you for the quick update! LGTM!

packages/eslint-plugin-svelte/src/rules/prefer-derived-over-derived-by.ts
```diff
@@ -0,0 +1,74 @@
...
+					expressionNode = arg.body.body[0].argument;
+				}
+				if (expressionNode === null) {
+					return;
+				}
+				context.report({
+					node,
+					messageId: 'unnecessary',
+					fix(fixer) {
+						const expressionText = context.sourceCode.getText(expressionNode!);
```
ota-meshi:
```suggestion
						const expressionText = context.sourceCode.getText(expressionNode);
```

PR #1532, opened 2026-05-23
feat(no-navigation-without-resolve): recognizing nullish TS types as allowed

### LINKED ISSUE #1525: no-navigation-without-resolve: false positive for links with optional ResolvedPathname href
### Description

The `no-navigation-without-resolve` rule reports a false positive on `<a>` elements when `href` is typed as an optional `ResolvedPathname` (i.e. `ResolvedPathname | undefined`).

The rule correctly passes `allowNullish: true` when checking link attributes, intending to permit nullish href values. However, when the href expression is not itself nullish but has a *type* that includes `undefined` (because the property is optional), neither `expressionIsNullish` nor `expressionIsResolvedPathname` matches:

- `expressionIsNullish` checks whether the expression is syntactically nullish (e.g. a literal `undefined`), not whether the type is nullable — so it returns `false` for `href`
- `expressionIsResolvedPathname` performs a bidirectional assignability check against `ResolvedPathname`, which fails for `ResolvedPathname | undefined` because `undefined` is not assignable to `ResolvedPathname`

### Reproduction

```svelte
<script lang="ts">
import type { ResolvedPathname } from '$app/types';

interface Props {
  href?: ResolvedPathname;
}

const { href }: Props = $props();
</script>

<a {href}>Click me!</a>
```

**Expected:** No error — `href` is either `ResolvedPathname` or `undefined`, both of which are valid for `<a href>`.

**Actual:** `linkWithoutResolve` is reported.

### Suggested fix

Strip nullability from the node type in `expressionIsResolvedPathname` before performing the bidirectional assignability check, using `checker.getNonNullableType()`. This way, `ResolvedPathname | undefined` is reduced to `ResolvedPathname`, which correctly passes the check.

See PR #1523.

marekdedic:
Closes #1525

Supersedes #1523

### DIFF
--- a/.changeset/silly-frogs-rush.md
+++ b/.changeset/silly-frogs-rush.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': minor
+---
+
+feat(no-navigation-without-resolve): recognizing nullish TS types as allowed
--- a/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
+++ b/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
@@ -317,7 +317,7 @@ function isValueAllowed(
 			variable.identifiers.length > 0 &&
 			variable.identifiers[0].parent.type === 'VariableDeclarator'
 		) {
-			if (expressionIsResolvedPathname(variable.identifiers[0], tsTools)) {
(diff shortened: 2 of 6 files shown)


ota-meshi (approved):
LGTM! Thanks!

PR #1545, opened 2026-05-30
feat: add no-at-const-tags rule

baseballyama:
## Summary

- Adds a new `svelte/no-at-const-tags` rule that reports legacy `{@const ...}` and autofixes it to the Svelte 5 `{const ...}` declaration tag.
- The rule is gated to Svelte 5 via `conditions: [{ svelteVersions: ['5'] }]` since `{const ...}` declaration tags were introduced in [Svelte 5.56.0](https://svelte.dev/docs/svelte/declaration-tags). On Svelte 3/4 the rule is inert.
- Per the [official docs](https://svelte.dev/docs/svelte/@const), `{@const}` is legacy syntax and `{let/const ...}` declaration tags are the recommended replacement.

## Test plan

- [x] `pnpm run mocha "tests/src/**/*.ts" -- -g no-at-const-tags` passes
- [x] Full `pnpm run test` passes (1433 tests)
- [x] `pnpm run lint:es` clean for the new files
- [x] Fixtures exercise both plain `{@const ...}` and the `{ @const ...}` whitespace variant in `{#if}` / `{#each}` / `{#snippet}` parents
- [x] Autofix output verified via the generated `*-output.svelte`


packages/eslint-plugin-svelte/src/rules/no-at-const-tags.ts
```diff
@@ -0,0 +1,64 @@
...
+		},
+		fixable: 'code',
+		schema: [],
+		messages: {
+			unexpected: 'Use `{const ...}` declaration tag instead of legacy `{@const ...}`.'
+		},
+		type: 'suggestion',
+		conditions: [
+			{
+				svelteVersions: ['5']
```
ota-meshi, line 23: I think this should be addressed in a separate pull request, but I think it would be convenient if we could check whether it's Svelte 5.56.0 or higher so that we can recommend this rule.

packages/eslint-plugin-svelte/tests/fixtures/rules/no-at-const-tags/invalid/const02-output.svelte
```diff
@@ -0,0 +1,13 @@
...
+	let items = [1, 2, 3];
+</script>
+
+{#if items.length}
+	{const total = items.reduce((a, b) => a + b, 0)}
+	Total: {total}
+{/if}
+
+{#snippet row(value)}
+	{ const doubled = value * 2 }
```
ota-meshi: I'm still not entirely sure about this, but is `$derived()` unnecessary here?

ota-meshi (commented):
Thank you for the PR! This rule is great!
I have a question.

packages/eslint-plugin-svelte/src/rules/no-at-const-tags.ts
baseballyama: I also think this extension is something we should do, since new features will likely continue to be added to Svelte 5 in the future.

packages/eslint-plugin-svelte/tests/fixtures/rules/no-at-const-tags/invalid/const02-output.svelte
baseballyama: ~~This test does not have `runes: true` in the config, so the mode is determined automatically by analyzing the Svelte file. Since the file does not contain any runes at all, it is correctly classified as a non-runes file. Therefore, I believe it is correct that `$derived` is not added.~~

--> However, I just realized that `{let}` and `{count}` can only be used in runes mode, so my earlier reasoning doesn’t actually hold.😇
Therefore, we cannot apply the rule to non-runes files.


https://svelte.dev/playground/hello-world?version=5.56.0#H4sIAAAAAAAAE22PwUoEMQyGXyXEi8qww3is0wVvvoP10JnJQKGbljbdVUrfXeqgXryFL_8X8ldkeyFU-EreB7iF5De4p80JbQ844O48ZVRvFeUz9lwHOPxYLzGe8pW8dLbYTP_xNbAQS0aF8wFViOICZ0iFKeu6W5-pwXg2bHjOa3JR-iyeBPYQQEOFNRQWUDBBezY8j38xrnduhwm01jC17tU1cBZYbALdD5wO-RGejvViUzNcR7c3w7PjWAQWx5u6Wl9I11-lQe-tDXK5LJQMfj-JAwp9CCpJhdr7gGKdvzneUB1VvgCOtGO-VwEAAA

baseballyama: updated: https://github.com/sveltejs/eslint-plugin-svelte/pull/1545/changes/8ea926a1aac18597e2cfd379850cb66d2a7d24ad

packages/eslint-plugin-svelte/src/rules/no-at-const-tags.ts
```diff
@@ -0,0 +1,65 @@
...
+						// initializer in `$derived(...)`. Skip when already wrapped.
+						if (
+							init.type === 'CallExpression' &&
+							init.callee.type === 'Identifier' &&
+							init.callee.name === '$derived'
+						) {
+							return;
+						}
+						yield fixer.insertTextBefore(init, '$derived(');
+						yield fixer.insertTextAfter(init, ')');
```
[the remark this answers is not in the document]
baseballyama: `declaration` is deprecated.

packages/eslint-plugin-svelte/tests/fixtures/rules/no-at-const-tags/valid/_requirements.json
```diff
@@ -0,0 +1,3 @@
+{
+	"svelte": ">=5.0.0"
+}
```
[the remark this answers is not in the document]
baseballyama: Since we currently test only against the latest version within each major release, this isn’t actually a problem in practice.

packages/eslint-plugin-svelte/tests/fixtures/rules/no-at-const-tags/invalid/_requirements.json
```diff
@@ -0,0 +1,3 @@
+{
+	"svelte": ">=5.0.0"
+}
```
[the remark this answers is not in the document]
baseballyama: Since we currently test only against the latest version within each major release, this isn’t actually a problem in practice.

docs/rules/no-at-const-tags.md
```diff
@@ -0,0 +1,59 @@
...
+
+## :book: Rule Details
+
+This rule reports uses of `{@const ...}` in runes mode.
+
+In Svelte 5, the `{@const ...}` tag is considered legacy. Use the `{const ...}` declaration tag instead, which can be placed anywhere inside the component.
+
+`{@const ...}` is reactive — its value is re-evaluated when its dependencies change. To preserve that behavior, the initializer must be wrapped in `$derived(...)`, and the auto-fix does this automatically.
+
+This rule does nothing in non-runes mode: it neither reports nor fixes there. `$derived(...)` is unavailable outside runes mode, so the reactive behavior of `{@const ...}` cannot be preserved when converting to a `{const ...}` declaration tag.
```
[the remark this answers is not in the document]
baseballyama: Apply part of it. https://github.com/sveltejs/eslint-plugin-svelte/pull/1545/changes/661df4048f5ac22c37d34d60d3418f9e7ebd9b10

ota-meshi (approved):
LGTM! Thank you!

PR #1557, opened 2026-06-26
feat: add `svelte/no-useless-props` rule

baseballyama:
## What

A new **opt-in** rule, `svelte/no-useless-props`, that flags component props which **no call site in the project ever passes** — as an attribute (`<C prop={…}>`), a binding (`bind:prop`), a spread, or body/`{#snippet}` content. Such a prop always takes its default: it is dead from the consumer side and should usually be removed (or a caller is missing).

It is the **whole-program counterpart** to [`no-unused-props`](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-unused-props/): that rule flags a prop never *read inside* its component (single-file); this one flags a prop never *passed from outside* (cross-file). Together they catch both ends of a dead prop.

## How

"Is this prop ever passed?" can't be answered from one file, so the rule does a **one-time whole-program pre-scan** the first time it runs, caches the result, and then reports per file. The analysis is powered by the optional [`svelte-shaker`](https://www.npmjs.com/package/svelte-shaker) dependency (the Svelte tree-shaker), loaded synchronously via `require(ESM)` (Node ≥ 22.12). If it isn't installed / Node is too old, the rule prints a one-time notice and no-ops.

Sound by construction (no false positives by design):

- components with **zero call sites** are skipped — entries / route pages whose props are framework-injected (a SvelteKit `+page.svelte`'s `data`, a `mount()` target);
- components that **escape as a value** (`<svelte:component this={C}>`, `const X = C`) are skipped — their prop usage is unknowable;
- a prop is reported only when **every** call site neither names it nor carries a spread that might set it;
- an incomplete scan (e.g. an unfollowed alias) can only **under**-report.

Targets Svelte 5 runes (`$props()`). Not in the recommended preset (opt-in).

### Options

- `include` (`string[]`, default `["src"]`) — dirs that contain the whole app (every call site must be inside them).
- `aliases` (`Record<string,string>`, default `{ "$lib": "src/lib" }`) — import-prefix → dir, so aliased imports resolve during the scan.

## Notes / for review

- `svelte-shaker` is intentionally **not** declared as a peer dependency (it would force the plugin's supported-Node range up via `compat-engines`); it is loaded optionally at runtime, the same pattern the plugin uses for `typescript`/`sass`/etc. It is a `devDependency` here for the rule's types and tests.
- The whole-program scan is cached per `(cwd, include, aliases)` for the lint run. Editor staleness (linting unsaved text against an on-disk scan) is guarded: a finding is only reported if its recorded span still spans the prop name.
- Tests drive the rule against real fixture files (`App.svelte` renders `Child` passing `used` but never `unused`) with absolute `include`/`filename`, and skip on Node < 22.12.

All 1429 tests pass; `tsc`, `eslint`, and `type-coverage` are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ota-meshi:
That new rule is fantastic!
Should we first drop support for older Node versions to make svelte-shaker usable without additional installation?

baseballyama:
@ota-meshi 

Sorry! While I was away having lunch, Claude Code went ahead and opened a PR on its own 😅

I haven’t finished reviewing the code yet, including whether the overall approach is the right one, so I’ll switch it back to a draft for now.

Yeah, it might be a good time to cut a new major release of `eslint-plugin-svelte` and officially drop support for older Node versions.
That said, I’d also like to work through the backlog of bug fixes first 😅

marekdedic:
Hi, just a drive-by comment: Doesn't this go explicitly against the design philosophy of eslint? (Linting is file-local). Also, what's the performance impact of this?

baseballyama:
I don’t think performance will be a significant issue. Parsing a Svelte project with 650 components takes only about 50 ms.

For VS Code, we plan to run svelte-shaker as a daemon, reducing the processing time to around 1.3 ms per file save.

From an ESLint architecture perspective, this is also not an issue. It’s essentially the same model as TypeScript ESLint, which requires analyzing multiple files to resolve type information.

PR #1559, opened 2026-07-14
feat: add `no-bind-value-on-checkable-inputs` rule

knrdl:
This rule prevents a common gotcha when it comes to HTML inputs and Svelte:

For most `<input>` types, the user-editable state is exposed through the `value` property. Checkboxes (and radios) are an exception: their state is represented by the `checked` property instead. While this distinction was useful and consistent back in the days of plain HTML forms, it can be confusing when using Svelte bindings.

Therefore this rule disallows `bind:value` on `<input type="checkbox">` and suggests the correct `bind:checked` instead.

Hoping to help preventing an easy-to-miss mistake and aligning checkbox bindings with Svelte's intended usage.


packages/eslint-plugin-svelte/src/rules/no-checkbox-bind-value.ts
```diff
@@ -0,0 +1,59 @@
+import type { AST } from 'svelte-eslint-parser';
+import { createRule } from '../utils/index.js';
+
+export default createRule('no-checkbox-bind-value', {
+	meta: {
+		fixable: 'code',
+		docs: {
+			description: 'disallow useless `bind:value` on `<input type="checkbox">`',
+			category: 'Possible Errors',
+			recommended: true
```
marekdedic: Based on my experience with adding new rules, I'd recommend letting it sit outside recommended for a while and only then adding it - not because it shouldn't (eventually) be recommended, just to play it safe

ota-meshi:
Thank you for this PR!
I have a question: are `<input type="radio">` elements excluded from this rule check? Or is `bind:value` valid for `<input type="radio">`?

knrdl:
You are right that it applies to radios, too. However, it is much harder to get it wrong with radios in Svelte because `value` has a semantic there:

```svelte
<script lang="ts">
    let checked = $state(true);

    let selection = $state('cat');
</script>

<!-- value is unused -->
<input type="checkbox" value="irrelevant" bind:checked />
{checked}

<!-- value is used -->
<input type="radio" value="cat" bind:group={selection} />
<input type="radio" value="dog" bind:group={selection} />
<input type="radio" value="fish" bind:group={selection} />
{selection}
```

A more strict rule would be:
* disallow `value` and `bind:value` on checkbox
* disallow `bind:value` on radio

A disadvantage might be that it would break if Svelte is used together with a library which imperatively manipulates the DOM like this:
```svelte
<script lang="ts">
    let selection = $state('cat');

    let felineType = $state<'cat' | 'lion'>('cat');

    function onclick() {
        document.getElementById('feline').value = 'lion';
    }
</script>

<input type="radio" id="feline" bind:value={felineType} bind:group={selection} />
{felineType}
<input type="radio" value="dog" bind:group={selection} />
<input type="radio" value="fish" bind:group={selection} />
{selection}

<button type="button" {onclick}>bigger cats!</button>
```
This is kind of an example of how not to do it but technically valid. Therefore I limited the rule to checkboxes.

ota-meshi:
Thank you for the comment.
I don't think we need to take into account the DOM manipulation you shared. Therefore I think it would be useful if the rule supported radio buttons as well.


> disallow `value` and `bind:value` on checkbox

We can use `value` with checkboxes as well.

```svelte
<input
	type="checkbox"
	name="flavours"
	value={flavour}
	bind:group={flavours}
/>
```

https://svelte.jp/tutorial/svelte/group-inputs

Could you please modify this PR?
Could you also consider a clearer name for the rule?

packages/eslint-plugin-svelte/src/rules/no-checkbox-bind-value.ts
knrdl: no problem, changed.

knrdl:
Good point, I hadn't thought about `bind:group` on checkboxes.

I've updated the rule and renamed it to `no-bind-value-on-checkable-inputs`. If you can think of a better name, please let me know.

packages/eslint-plugin-svelte/src/rules/no-bind-value-on-checkable-inputs.ts
```diff
@@ -0,0 +1,75 @@
...
+			category: 'Possible Errors',
+			recommended: false
+		},
+		schema: [
+			{
+				type: 'object',
+				properties: {},
+				additionalProperties: false
+			}
+		],
```
ota-meshi: Since no options are provided, the schema must be empty.


```suggestion
		schema: [],
```

packages/eslint-plugin-svelte/src/rules/no-bind-value-on-checkable-inputs.ts
```diff
@@ -0,0 +1,75 @@
...
+								((attr.value[0].type === 'SvelteLiteral' &&
+									attr.value[0].value.toLowerCase() === inputType) || // type="checkbox"
+									(attr.value[0].type === 'SvelteMustacheTag' &&
+										attr.value[0].kind === 'text' &&
+										attr.value[0].expression.type === 'Literal' &&
+										typeof attr.value[0].expression.value === 'string' &&
+										attr.value[0].expression.value.toLowerCase() === inputType)) // type={"checkbox"}
+						)
+					);
+				}
```
ota-meshi: I think we can resolve this in a single call by returning a string and utilizing it.
Also, could you please modify the code to break down the conditional statements and perform checks in stages, while making use of utility functions?

```js
import { getStaticValue } from '@eslint-community/eslint-utils';
import { getScope } from '../utils/ast-utils.js';
```

```suggestion
				function getType(): string | null {
					const typeAttr = node.startTag?.attributes.find(
						(attr): attr is AST.SvelteAttribute =>
							attr.type === 'SvelteAttribute' && attr.key.name === 'type'
					);
					if (!typeAttr) return null;
					if (typeAttr.value.length !== 1) return null;
					const typeValue = typeAttr.value[0];
					if (typeValue.type === 'SvelteLiteral') {
						// <input type="checkbox" />
						return typeValue.value.toLowerCase();
					}
					if (typeValue.type === 'SvelteMustacheTag') {
						const staticValue = getStaticValue(
							typeValue.expression,
							getScope(context, typeValue.expression)
						);
						if (typeof staticValue?.value !== 'string') return null
						 // <input type={"checkbox"} />;
						return staticValue.value.toLowerCase()
					}
					return null;
				}
```

packages/eslint-plugin-svelte/src/rules/no-bind-value-on-checkable-inputs.ts
```diff
@@ -0,0 +1,75 @@
...
+
+				const isCheckbox = getType('checkbox');
+				const isRadio = getType('radio');
```
ota-meshi:
```suggestion
				const type = getType();
				if (!type) return;
				const isCheckbox = type === 'checkbox';
				const isRadio = type === 'radio';
```

packages/eslint-plugin-svelte/src/rules/no-bind-value-on-checkable-inputs.ts
```diff
@@ -0,0 +1,75 @@
...
+				if (bindValue && (isCheckbox || isRadio)) {
+					context.report({
+						node: bindValue,
+						messageId: isCheckbox ? 'checkbox' : 'radio',
+						fix(fixer) {
+							if (isCheckbox) {
+								if (bindValue.shorthand)
+									return fixer.replaceText(bindValue, 'bind:checked={value}');
+								return fixer.replaceText(bindValue.key, 'bind:checked');
+							}
```
ota-meshi: I'm unsure whether it's appropriate to apply an auto-fix for checkboxes; the user might actually want to use `bind:group`.

Instead, how about providing a fix suggestion via the suggestion API for checkboxes?

```js
suggest: [
  {...}
]
```

What do you think?

ota-meshi (changes requested):
Thank you for updating the PR!
The rule name you came up with is far better than what I had imagined.

packages/eslint-plugin-svelte/src/rules/no-bind-value-on-checkable-inputs.ts
knrdl: Offering a suggestion instead of a auto-fix is definitely the safer choice as the user’s mental model is already off track at that point. What do you think of suggesting `bind:checked` and `bind:group` for checkboxes, and `bind:group` for radios?

knrdl:
I've updated the PR, please have a look

I thought a bit about non-lowercase tag and attribute names:
* `<iNPUT bind:value>` already raises https://svelte.dev/docs/svelte/compiler-errors#bind_invalid_target 
* `<input bind:VALUE>` already raises https://svelte.dev/docs/svelte/compiler-errors#bind_invalid_name
* `input type="CHECKBOX">` already covered.
* `<input TYPE="checkbox">` appears to be completely valid, so I added coverage for that case.

packages/eslint-plugin-svelte/src/rules/no-bind-value-on-checkable-inputs.ts
ota-meshi: > What do you think of suggesting bind:checked and bind:group for checkboxes, and bind:group for radios?

I think it would be even more convenient if that suggested option you mentioned were available!

ota-meshi (approved):
LGTM! Thank you!

PR #1563, opened 2026-07-20
feat: add `no-conflicting-module-names` rule

### LINKED ISSUE #1562: New rule: `no-conflicting-module-names` (disallow `Foo.svelte.ts` next to `Foo.svelte`)
### Motivation

Since Svelte 5, `.svelte.ts` / `.svelte.js` files (runes modules) are a normal part of a Svelte project. A runes module can have the same base name as a component:

```
src/
├── Foo.svelte      # a component
└── Foo.svelte.ts   # a runes module (a different module)
```

This layout is legal today, and at runtime there is no ambiguity. But the import specifier `./Foo.svelte` resolves to different modules depending on the tool:

| Resolver | `./Foo.svelte` resolves to |
| --- | --- |
| Vite (runtime / bundler) | `Foo.svelte` (the component) |
| svelte-check / svelte-language-tools (editors) | `Foo.svelte` (the component) |
| Plain TypeScript resolution (`tsc`, typescript-eslint type-aware lint) | `Foo.svelte.ts` (the module) |

TypeScript does not know the `.svelte` extension, so it appends `.ts` and finds `Foo.svelte.ts` first. So tools that use plain TypeScript resolution type the component import as the sibling module, which is not what actually runs. The editor and `svelte-check` patch module resolution, so they look fine. This makes the problem hard to notice.

Runnable repro: https://github.com/baseballyama/svelte-module-collision-repro (`npm install && npm run demo`)

I think this is a real problem, not a theoretical one:

- svelte-language-tools used `.svelte.ts` as its virtual file suffix before. After Svelte 5 made `.svelte.ts` a real extension, they had to move to `.d.svelte.ts`. So this name collision already happened once in the ecosystem.
- The experimental importer prop typing in svelte-eslint-parser (sveltejs/svelte-eslint-parser#913) resolves `.svelte` imports with plain TS resolution, through a virtual `X.svelte.ts` file. If a real `X.svelte.ts` exists, the real file must win, and the parser cannot provide component prop types for that component. A lint rule that prevents the collision would help here.

Also, the failure is silent in both cases. With the sibling file, the import is typed as the module. Without it, TS falls back to Svelte's ambient `declare module '*.svelte'` (`any` props). In both cases there is no error that points at the real problem.

### Description

I propose a new rule (working name: `svelte/no-conflicting-module-names`). It reports a `.svelte` file when a sibling file with the same base name and a conflicting extension exists:

- `Foo.svelte` + `Foo.svelte.ts` / `.tsx` / `.js` / `.jsx` / `.mts` / `.cts` / `.mjs` / `.cjs` → report

Details:

- The rule checks the filesystem for sibling files when linting the `.svelte` file (similar to other fs-based rules like `import/no-unresolved`). It reports once per file. The message explains the resolution problem and suggests a rename (for example `foo-state.svelte.ts`).
- No autofix. Renaming a module also requires updating its imports.
- I suggest starting as opt-in (not in `recommended`). The layout is legal today and only breaks tools that use plain TS resolution.

### Examples

This rule checks the file layout, not code patterns:

```
✓ GOOD
src/
├── Foo.svelte
└── foo-state.svelte.ts

✗ BAD
src/
├── Foo.svelte
└── Foo.svelte.ts    # `./Foo.svelte` resolves here for plain TS resolution
```

### Additional comments

- Declaration files are not in the default target list. A hand-written `Foo.svelte.d.ts` is a valid way to type a component (svelte-language-tools prefers it over the component itself). `Foo.d.svelte.ts` is the virtual file name used by language-tools. We could add an option to report them too.
- Open questions: the rule name, and whether the rule should also report the module file itself. Sometimes only one of the two files is included in the lint run.
- I am also thinking about a one-time warning in svelte-eslint-parser when its hook detects this collision. But a lint rule is still useful because it works no matter which tool the developer runs.

baseballyama:
close #1562


docs/rules/no-conflicting-module-names.md
```diff
@@ -0,0 +1,64 @@
...
+The following module extensions trigger a report: `.ts`, `.tsx`, `.js`, `.jsx`, `.mts`, `.cts`, `.mjs`, `.cjs`.
+
+Declaration files are not reported. `Foo.svelte.d.ts` and `Foo.d.svelte.ts` are legitimate and do not shadow the component.
+
+<!-- prettier-ignore-start -->
+```
+
+src/
+  Foo.svelte
+  Foo.svelte.ts   // ✗ BAD: collides with Foo.svelte
```
ota-meshi: The documentation marks `Foo.svelte.ts` as BAD, but I believe the report would actually be issued against `Foo.svelte`.

If you want to align with the documentation, wouldn't it be better to report the issue against `Foo.svelte.ts`? Alternatively, reporting it for both files might also be a good idea.

ota-meshi (changes requested):
Thank you for this PR!

docs/rules/no-conflicting-module-names.md
baseballyama: I updated it to report both `.svelte` and `.svelte.ts/.js`!

ota-meshi (approved):
LGTM! Thank you!

PR #1565, opened 2026-07-25
feat: report useless template literal mustaches

### LINKED ISSUE #893: Add rule: disable es6 template literals (eg ```{`foo${bar}`}```)  in attributes
### Motivation

some developers (and ai, too!) will write attributes in jsx style. eg ```<a href={`https://example.com${path}`}>link</a>```.

in idiomatic svelte, this should be `<a href="https://example.com{path}">link</a>`

### Description

warn on ```attr={``}```

### Examples

```svelte
<script>
</script>

<!-- ✓ GOOD -->
<Foo attr="foo${bar}" />

<!-- ✗ BAD -->
<Foo attr={`foo${bar}`} />
```

### Additional comments

_No response_

ColumbusLabs:
Closes #893.

## Summary
- extend svelte/no-useless-mustaches to report dynamic template literals used as an entire attribute value
- keep embedded, conditional, style-directive, comment, escape, multiline, and raw-brace cases exempt
- add fixture coverage, documentation, and a changeset

The new diagnostic is intentionally report-only because a safe conversion must account for Svelte attribute escaping.

## Validation
- focused rule suite: 12 passing
- full workspace suite: 1,455 passing
- package build and docs production build
- targeted ESLint, docs lint, Prettier, and git diff --check

### DIFF
--- a/packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-attribute-interpolation/invalid/template-literal-attribute-output.svelte
+++ b/packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-attribute-interpolation/invalid/template-literal-attribute-output.svelte
@@ -1,0 +1,9 @@
+<script>
+	let foo = 'foo';
+	let bar = 'bar';
+</script>
+
+<Foo attr={`prefix${foo}`} />
+<div data-text={`prefix${foo}${bar}`} />
+<!-- prettier-ignore -->
+<Foo attr="{`prefix${foo}`}" />
--- a/packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-attribute-interpolation/invalid/template-literal-attribute-output.svelte
+++ b/packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-attribute-interpolation/invalid/template-literal-attribute-output.svelte
@@ -1,9 +1,0 @@
-<script>
-	let foo = 'foo';
-	let bar = 'bar';
-</script>
-
-<Foo attr={`prefix${foo}`} />
-<div data-text={`prefix${foo}${bar}`} />
-<!-- prettier-ignore -->
-<Foo attr="{`prefix${foo}`}" />
--- a/.changeset/tidy-mustaches-smile.md
+++ b/.changeset/tidy-mustaches-smile.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': minor
+---
+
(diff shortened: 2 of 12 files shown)


ota-meshi (changes requested):
Thank you for this PR!

However, I think it would be better to provide this via a separate rule instead of modifying `no-useless-mustaches`.
What do you think?

ColumbusLabs:
Agreed—a separate rule keeps `no-useless-mustaches`’ existing recommended behavior stable. I propose `svelte/prefer-attribute-interpolation`: report only an attribute whose entire value is a dynamic template literal, while leaving embedded, conditional, style-directive, comment, escape, multiline, and raw-brace cases alone. Does that name and boundary match what you had in mind? If so, I’ll refactor this PR in place.

ota-meshi:
That rule sounds good to me!
It might be even more useful if it could also check string concatenation using `+`, but that's not strictly necessary as it can be achieved by combining it with the `prefer-template` rule.
https://eslint.org/docs/latest/rules/prefer-template

ColumbusLabs:
Implemented this as the separate `svelte/prefer-attribute-interpolation` rule and removed the change from `no-useless-mustaches`. It reports only dynamic template literals used as an attribute’s entire value; `+` concatenation is intentionally left to ESLint’s `prefer-template` rule. Added dedicated docs and fixtures. Verified with the focused rule tests, 1,456 package tests, TypeScript build, ESLint, docs build/lint, and Prettier.

packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-attribute-interpolation/invalid/template-literal-attribute-output.svelte
ota-meshi, line 1: It appears that the rule do not support auto-fixing.
Could you please delete this file or add auto-fix support to the rule?

ota-meshi (changes requested):
Thank you for implementing the rule!

packages/eslint-plugin-svelte/tests/fixtures/rules/prefer-attribute-interpolation/invalid/template-literal-attribute-output.svelte
ColumbusLabs: Deleted this non-fixable output fixture in a6c49254. The rule remains report-only with no autofix; focused and package verification passed.

ota-meshi (approved):
LGTM! Thank you!

PR #1566, opened 2026-08-01
fix(no-unused-props): treat an object spread as consuming every nested property

### LINKED ISSUE #1556: False positive `svelte/no-unused-props` when object prop property is used via spread in script

9.x

3.19.0

### What did you do?

```svelte
<!-- App.svelte -->
<script lang="ts">
  type Props = {
    app: { name: string; description: string };
  };
  const { app }: Props = $props();

  const data = { ...app, name: `${app.name} (copy)` };
</script>

<p>{data.name}</p>
```

No error. `app.description` is used because `app` is spread into a new object — all properties of `app` (including `description`) flow into `data`.

### What actually happened?

```
'description' in 'app' is an unused property  svelte/no-unused-props
```

### Additional context

This is a different variant from the spread-into-component case fixed by #1378 (which handled `<Child {...props} />`). Here the prop is an object whose properties are consumed via a JavaScript object spread expression in `<script>`, not a component spread in the template.

The rule tracks which specific properties of an object-typed prop are accessed (e.g. `app.name`), but does not recognize that spreading the object (`{ ...app }`) constitutes usage of all its properties.

arpitjain099:
Fixes #1556.

Spreading a prop into an object literal consumes all of its properties, but the rule still reported the ones that were not also accessed by name:

```svelte
<script lang="ts">
	type Props = {
		app: { name: string; description: string };
	};
	const { app }: Props = $props();

	const data = { ...app, name: `${app.name} (copy)` };
</script>

<p>{data.name}</p>
```

gives `'description' in 'app' is an unused property`.

The spread itself is already recorded: `getPropertyPath` sets `isSpread` for a `SpreadElement` parent, so `usedSpreadPropertyPaths` contains `app`. The problem is downstream in `checkUnusedProperties`:

```js
const isUsedThisInPath = usedPropertyPaths.includes(currentPathStr) || usedSpreadPropertyPaths.some(...);
const isUsedInPath = usedPropertyPaths.some((path) => path.startsWith(`${currentPathStr}.`));

if (isUsedThisInPath && !isUsedInPath) {
	continue;
}
```

For `app` both are true, since `app.name` is a direct access, so the early continue does not fire and the walk descends into `app` and reports `description`. In other words the spread coverage is silently dropped as soon as any nested property is also read by name. Removing `${app.name}` from the example makes the report go away, which is what made this look like a spread-detection problem rather than a precedence one.

This adds a check before that block: if a spread covers this exact path, or any ancestor of it, skip the property entirely, since everything beneath it has flowed into the spread target.

Existing behavior is unchanged elsewhere. The `path === ''` case (a root `$props()` spread) and the "spread deeper inside marks the parent used" case both still go through `isUsedThisInPath`.

Test is a valid fixture, `spread-object-expression-input.svelte`, holding the snippet from the issue. It reports one error before the change and none after. Full suite is 1455 passing.

### DIFF
--- a/.changeset/clever-pandas-shave.md
+++ b/.changeset/clever-pandas-shave.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-svelte': patch
+---
+
+fix: `svelte/no-unused-props` false positive when an object prop is consumed via an object spread and also accessed directly
--- a/.changeset/clever-pandas-shave.md
+++ b/.changeset/clever-pandas-shave.md
@@ -2,4 +2,4 @@
 'eslint-plugin-svelte': patch
 ---
 
-fix: `svelte/no-unused-props` false positive when an object prop is consumed via an object spread and also accessed directly
+fix: `svelte/no-unused-props` false positive when a prop is consumed via an object spread and also accessed by name
--- a/packages/eslint-plugin-svelte/src/rules/no-unused-props.ts
+++ b/packages/eslint-plugin-svelte/src/rules/no-unused-props.ts
@@ -315,21 +315,20 @@ export default createRule('no-unused-props', {
 
 				if (reportedPropertyPaths.has(currentPathStr)) continue;
 
-				const propType = typeChecker.getTypeOfSymbol(prop);
-
-				// A spread of this property, or of anything containing it, consumes
-				// every property beneath it, so there is nothing left to report
-				// even when some of them are also accessed individually.
-				const isSpreadInThisPath = usedSpreadPropertyPaths.some((path) => {
+				// Spreading this property, or any of its ancestors, consumes everything
+				// beneath it, so nothing under it can be reported -- even when some of
+				// those properties are also read by name.
+				const isCoveredBySpread = usedSpreadPropertyPaths.some((path) => {
 					return path === '' || path === currentPathStr || currentPathStr.startsWith(`${path}.`);
 				});
-				if (isSpreadInThisPath) continue;
+				if (isCoveredBySpread) continue;
+
+				const propType = typeChecker.getTypeOfSymbol(prop);
 
 				const isUsedThisInPath =
 					usedPropertyPaths.includes(currentPathStr) ||
-					usedSpreadPropertyPaths.some((path) => {
-						return path === '' || path === currentPathStr || path.startsWith(`${currentPathStr}.`);
-					});
+					// A spread deeper inside this property counts as using the property itself.
+					usedSpreadPropertyPaths.some((path) => path.startsWith(`${currentPathStr}.`));
 				const isUsedInPath = usedPropertyPaths.some((path) => {
 					return path.startsWith(`${currentPathStr}.`);
 				});
--- a/packages/eslint-plugin-svelte/src/rules/no-unused-props.ts
+++ b/packages/eslint-plugin-svelte/src/rules/no-unused-props.ts
@@ -317,6 +317,14 @@ export default createRule('no-unused-props', {
 
 				const propType = typeChecker.getTypeOfSymbol(prop);
 
+				// A spread of this property, or of anything containing it, consumes
+				// every property beneath it, so there is nothing left to report
+				// even when some of them are also accessed individually.
+				const isSpreadInThisPath = usedSpreadPropertyPaths.some((path) => {
+					return path === '' || path === currentPathStr || currentPathStr.startsWith(`${path}.`);
+				});
+				if (isSpreadInThisPath) continue;
+
 				const isUsedThisInPath =
 					usedPropertyPaths.includes(currentPathStr) ||
 					usedSpreadPropertyPaths.some((path) => {
--- a/packages/eslint-plugin-svelte/tests/fixtures/rules/no-unused-props/valid/spread-object-expression-input.svelte
+++ b/packages/eslint-plugin-svelte/tests/fixtures/rules/no-unused-props/valid/spread-object-expression-input.svelte
@@ -1,0 +1,11 @@
+<!--Test1.svelte-->
+<script lang="ts">
+	type Props = {
+		app: { name: string; description: string };
+	};
+	const { app }: Props = $props();
+
+	const data = { ...app, name: `${app.name} (copy)` };
+</script>
+
+<p>{data.name}</p>
--- a/packages/eslint-plugin-svelte/tests/fixtures/rules/no-unused-props/invalid/spread-sibling-unused-errors.yaml
+++ b/packages/eslint-plugin-svelte/tests/fixtures/rules/no-unused-props/invalid/spread-sibling-unused-errors.yaml
@@ -1,0 +1,4 @@
+- message: "'z' in 'b' is an unused property."
+  line: 8
+  column: 6
+  suggestions: null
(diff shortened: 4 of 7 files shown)


packages/eslint-plugin-svelte/src/rules/no-unused-props.ts
```diff
@@ -317,6 +317,14 @@ export default createRule('no-unused-props', {
 
 				const propType = typeChecker.getTypeOfSymbol(prop);
 
+				// A spread of this property, or of anything containing it, consumes
+				// every property beneath it, so there is nothing left to report
+				// even when some of them are also accessed individually.
+				const isSpreadInThisPath = usedSpreadPropertyPaths.some((path) => {
+					return path === '' || path === currentPathStr || currentPathStr.startsWith(`${path}.`);
+				});
+				if (isSpreadInThisPath) continue;
```
Stadly: The new early `continue` makes part of `isUsedThisInPath` below unreachable: if `path === ''` or `path === currentPathStr`, this check has already fired, so on line 331 only `path.startsWith(`${currentPathStr}.`)` is still live. I confirmed by deleting the other two clauses — all 1455 tests still pass.

Worth folding the two spread checks together so the next reader doesn't have to work out which half is load-bearing. This also hoists the check above `getTypeOfSymbol`, so the skipped case doesn't pay for the type lookup:

```suggestion
				// Spreading this property, or any of its ancestors, consumes everything
				// beneath it, so nothing under it can be reported -- even when some of
				// those properties are also read by name.
				const isCoveredBySpread = usedSpreadPropertyPaths.some((path) => {
					return path === '' || path === currentPathStr || currentPathStr.startsWith(`${path}.`);
				});
				if (isCoveredBySpread) continue;

				const propType = typeChecker.getTypeOfSymbol(prop);
```

Line 331 is outside the diff so I can't suggest it here, but it wants the same treatment:

```js
const isUsedThisInPath =
	usedPropertyPaths.includes(currentPathStr) ||
	// A spread deeper inside this property counts as using the property itself.
	usedSpreadPropertyPaths.some((path) => path.startsWith(`${currentPathStr}.`));
```

packages/eslint-plugin-svelte/tests/fixtures/rules/no-unused-props/valid/spread-object-expression-input.svelte
```diff
@@ -0,0 +1,11 @@
...
+<script lang="ts">
+	type Props = {
+		app: { name: string; description: string };
+	};
+	const { app }: Props = $props();
+
+	const data = { ...app, name: `${app.name} (copy)` };
+</script>
+
+<p>{data.name}</p>
```
Stadly, line 11: The risk this change carries is over-suppression, and there are currently **zero invalid spread fixtures in the entire rule** — so nothing guards that direction. A future refactor that dropped the `.` from the prefix check would sail through CI.

Three more fixtures cover it. I ran these against the branch: 1458 passing.

`valid/spread-nested6-input.svelte` — the nested variant this also fixes:

```svelte
<!--Test1.svelte-->
<script lang="ts">
	interface Props {
		a: {
			b: { c: string; d: number };
		};
	}

	let { a }: Props = $props();

	const data = { ...a.b, c: a.b.c };
</script>

<p>{data.c}</p>
```

`valid/spread-prefix-collision-input.svelte` — spread `a` must not suppress `ab.x`:

```svelte
<!--Test1.svelte-->
<script lang="ts">
	interface Props {
		a: { z: string };
		ab: { x: string };
	}

	let { a, ab }: Props = $props();

	const data = { ...a, z: a.z };
</script>

<p>{data.z}{ab.x}</p>
```

`invalid/spread-sibling-unused-input.svelte` — an unused sibling is still reported next to a spread:

```svelte
<!--Test1.svelte-->
<script lang="ts">
	interface Props {
		a: { x: string; y: string };
		b: { z: string };
	}

	let { a, b }: Props = $props();

	const data = { ...a, x: a.x };
</script>

<p>{data.x}</p>
```

with `invalid/spread-sibling-unused-errors.yaml`:

```yaml
- message: "'z' in 'b' is an unused property."
  line: 8
  column: 6
  suggestions: null
```

Stadly (commented):
Nice diagnosis — framing this as a precedence bug rather than a spread-detection bug is right, and it matches what I see in the code. `getPropertyPath` already records the spread; the coverage just gets dropped in `checkUnusedProperties` as soon as `isUsedInPath` is also true, so the walk descends and reports the siblings.

I checked this out and ran it against some probe fixtures. Baseline produced five false positives, and the patch clears all five while leaving the true positives intact:

| case | before | after |
| --- | --- | --- |
| `{...app}` + `app.name` (this issue) | `'description' in 'app'` | clean |
| nested `{...a.b}` + `a.b.c` | `'d' in 'a.b'` | clean |
| spread prop whose child is itself an object | `'meta' in 'app'` | clean |
| sibling prop `b` never used | `'z' in 'b'` | `'z' in 'b'` |
| prefix collision: spread `a`, unused `ab.y` | `'y' in 'ab'` | `'y' in 'ab'` |

So it fixes strictly more than the reported case — the nested variant comes along for free — and the `${path}.` suffix in the new check correctly keeps `a` from swallowing `ab`. I couldn't construct an over-suppression regression. Full suite is 1455 passing here too.

Two things I'd like to see before this lands, both inline.

One note on the PR description: it says the `path === ''` case still goes through `isUsedThisInPath`, but the new check short-circuits it first. The behavior is still correct — a root `$props()` spread does consume everything — the stated reasoning just doesn't match the code anymore.

And one thing worth stating out loud for the record, though I don't think it should block: this locks in treating a spread as consuming *every* property, so `const data = { ...app }` followed by only `data.name` will no longer flag `description`. That's a deliberate false-negative trade to kill the false positive, and it's consistent with how `spread-nested3` already behaves for `<Test {...props.a} />` — just noting it's now load-bearing in the script case too.

.changeset/clever-pandas-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-svelte': patch
+---
+
+fix: `svelte/no-unused-props` false positive when an object prop is consumed via an object spread and also accessed directly
```
Stadly: Minor: the fix covers more than this wording suggests — the nested case (`{...a.b}` alongside `a.b.c`) is fixed too, not just top-level object props. Maybe:

```suggestion
fix: `svelte/no-unused-props` false positive when a prop is consumed via an object spread and also accessed by name
```

arpitjain099:
All applied, thanks for writing it out so precisely.

Took the `isCoveredBySpread` block verbatim and hoisted it above `getTypeOfSymbol`. You were right that the early `continue` made the other two clauses of `isUsedThisInPath` dead: with your version, which drops them, the suite is 1458 passing here too, matching what you got.

All three fixtures are in, including `invalid/spread-sibling-unused` with your errors.yaml. Your point about there being zero invalid spread fixtures in the rule is the part that convinced me: nothing was guarding the over-suppression direction, so a later refactor dropping the `.` from the prefix check would have gone straight through CI.

Changeset reworded to your suggestion, since the nested case is fixed too and the old wording only mentioned top level props.

`prettier --check` clean on the touched files.

Stadly (approved):

arpitjain099:
@Stadly please merge this PR when you can

Stadly:
Would love to, but unfortunately I'm not a maintainer. So we'll have to wait for someone else to take a look.

PR #1579, opened 2026-09-10
feat(no-navigation-without-resolve): allow URL-derived hrefs

### LINKED ISSUE #1487: `no-navigation-without-resolve`: Allow URL.href

latest

master

### What did you do?

```svelte
<script>
  const one = new URL("https://svelte.dev");
  const two = new URL("/blog", "https://svelte.dev");
</script>

<a href={one.href}>One</a>
<a href={two.href}>Two</a>
<a href={one.toString()}>One</a>
<a href={two.toString()}>Two</a>
<a href={one}>One</a> <!-- Not 100% sure this calls toString -->
<a href={two}>Two</a> <!-- Not 100% sure this calls toString -->
```

No error

### What actually happened?

Error

Will make a PR

### Additional comments

_No response_

marekdedic:
Closes #1487

### DIFF
--- a/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
+++ b/packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
@@ -448,10 +448,78 @@ function expressionIsNullish(
 	}
 }
 
+function isGlobalIdentifier(
+	ctx: FindVariableContext,
+	node: TSESTree.CallExpressionArgument | TSESTree.Expression | AST.SvelteLiteral,
+	name: string
+): boolean {
+	if (node.type !== 'Identifier' || node.name !== name) {
+		return false;
+	}
+	const variable = ctx.findVariable(node);
+	return variable === null || variable.defs.length === 0;
+}
+
+function expressionIsUrl(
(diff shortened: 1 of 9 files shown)


packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
```diff
@@ -448,10 +448,78 @@ function expressionIsNullish(
 	}
 }
 
+function isGlobalIdentifier(
```
ota-meshi: I think we can use `ReferenceTracker`'s `iterateGlobalReferences` instead.
It is useful because it also tracks things like `globalThis.URL`.
https://eslint-community.github.io/eslint-utils/api/scope-utils.html#tracker-iterateglobalreferences

ota-meshi (changes requested):
Thank you for this PR!

packages/eslint-plugin-svelte/src/rules/no-navigation-without-resolve.ts
marekdedic: Good catch, fixed and added tests for it as well :)

PR #1581, opened 2026-09-23
fix(prefer-svelte-reactivity): don't report in non-runes components

### LINKED ISSUE #1580: `prefer-svelte-reactivity` reports in non-runes components, where `SvelteSet` & co. are not reactive

9.x

3.23.0 (svelte 5.57.1)

### What did you do?

A legacy-mode component (no runes) that keeps a plain `Set` as a non-reactive cache:

```svelte
<script>
	const seen = new Set();
	let count = 0;

	function visit(id) {
		if (seen.has(id)) return;
		seen.add(id);
		count += 1;
	}
</script>

<button on:click={() => visit('a')}>{count}</button>
```

No report. This component is in legacy mode, and following the suggestion does not make anything reactive there.

### What actually happened?

```
2:18  error  Found a mutable instance of the built-in Set class. Use SvelteSet instead  svelte/prefer-svelte-reactivity
```

https://github.com/badreddinesaadioui/prefer-svelte-reactivity-legacy-repro

`npx eslint Lint.svelte` shows the report, and `node --conditions=browser run.mjs` shows `SvelteSet` not updating in the legacy component.

#### Why the suggestion does not apply in legacy mode

In a non-runes component the compiler does not track `svelte/reactivity` classes. Compiling a legacy component that reads a `SvelteSet` in the template and in a `$:` statement gives:

```js
$.legacy_pre_effect(() => {}, () => {
	$.set(viaDollar, s.size);
});
// ...
$.template_effect(() => {
	$.set_text(text, ($.untrack(() => s.size)));
	// ...
});
```

The template read is wrapped in `$.untrack` and the `$:` statement has no dependencies, so `s.add(...)` updates neither. At runtime (`run.mjs` in the repro), after two `add()` calls the legacy component still renders `0` in both places, while the same component in runes mode renders `2`.

So in legacy mode the rule points you at a fix that does nothing, and the only way to silence it is to rewrite working code (I ended up replacing a `Set` with an array).

#### Proposed fix

Gate the rule on runes mode, the same way `prefer-writable-derived` already does:

```ts
conditions: [
	{
		svelteVersions: ['5'],
		svelteFileTypes: ['.svelte', '.svelte.[js|ts]'],
		runes: [true, 'undetermined']
	}
]
```

`.svelte.js/.ts` files are always runes, so they are unaffected. Components that are runes through a rune, `<svelte:options runes />` or `compilerOptions.runes` are still reported.

This is not the same as #1271: that issue covers runes components where the instance never escapes a function. This one covers legacy components, where the suggested classes are never reactive at all.

I have the fix and tests ready and can open a PR.

badreddinesaadioui:
In a non-runes component the compiler does not track `svelte/reactivity` classes (template reads are wrapped in `$.untrack`, `$:` statements get no dependencies), so telling the user to switch to `SvelteSet`/`SvelteMap`/`SvelteDate`/`SvelteURL` changes nothing. The rule now carries the same `runes: [true, 'undetermined']` condition as `prefer-writable-derived`.

Fixes #1580

**Tests**

- The existing `invalid/{date,map,set,url,url-search-params}` fixtures contain no runes, so the parser correctly sees them as legacy components. Each directory gets a `_config.json` with `svelteFeatures.runes: true`, so they keep testing what they were written for.
- New `valid/legacy-{set,map,date,url}01` fixtures: legacy components that must not be reported. All four fail on `main` and pass with this change.
- New `invalid/runes-option/set01`: a component that opts into runes with `<svelte:options runes />` is still reported.

`pnpm test` (1461 passing), `pnpm build`, `pnpm lint`, `pnpm typecov` and `pnpm update` (no generated changes) all pass locally.


marekdedic (approved):
LGTM, thanks!

